PR #5, opened 2026-08-08
docs: add full architecture/decision review and fix ADR index

Szczepanov:
Adds docs/analysis/2026-08-08-architecture-review.md -- a repository-wide
review covering the Python ingestion package, the TypeScript engine, all
nine ADRs, architecture docs, agent guidance and CI.

Headline findings (each verified against the working tree, two confirmed
with throwaway probes):

- UserContext.constraints.injuries is hardcoded empty by the only
  production adapter, so ADR-0007's documented hard injury gate is a
  no-op on both selection paths and across the whole forecast.
- Garmin-measured sessions attach an all-zero stimulus profile, which
  routes them to the vector-coverage credit path and earns them zero
  microcycle credit -- while *unrecognised* activity types fall through
  to the keyword fallback and do get credit.
- RecentHistoryEntry carries no date, so anti-stacking treats array
  adjacency as calendar adjacency, contradicting ADR-0008 section 6; the
  rolling arm also suppresses the third same-modality session in a week
  to 0.15x, which fights the 1.40x A-event boost.
- rules.ts and planner.ts call the optimizer with different context
  (no systemicCost, fabricated preferences), so the intensity-stacking
  guard is live in the forecast but dead for today's prescription.
- Home.tsx hardcodes the green next-day branch; ADR-0008 specifies the
  user-selected branch, and no selector exists.
- daily_recommendations rules pin only createdAt and allow a v3 document
  to be downgraded to v1, dropping its audit -- despite the inline
  comment describing the document as immutable audit evidence.

Also documents three parallel objective-credit models (one dead), the
half-finished stimulus-axis rename, missing ADRs for the weekly-anchor
and decision-provenance layers, a frozen POLICY_VERSION, an engine
architecture doc describing a mode hierarchy that no longer exists, and
a frontend test suite that cannot run without Firebase env vars.

Closes with a six-phase remediation plan ordered by consequence.

Also fixes docs/README.md: ADR-0009 was missing from the index.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01HNpyhhjdSgRmLkidyLEdq7

## Summary by CodeRabbit

- **Documentation**
  - Expanded the documentation index with new architecture decisions, reviews, and implementation plans.
  - Updated Recommendation Engine documentation with selection paths, scoring, modes, ranking, and multi-day planning.
  - Added guidance for decision provenance, audit replay, weekly planning anchors, and architecture governance.
  - Published phased plans covering safety, planning intent, ranking consistency, objective credit, fatigue modeling, testing, and sequence planning.
  - Clarified plan statuses, dependencies, acceptance criteria, historical references, and implementation standards.

### DIFF
--- a/docs/plans/phase-0-instrumentation.md
+++ b/docs/plans/phase-0-instrumentation.md
@@ -208,7 +208,8 @@ run surfaces, or record explicit per-module ignores with a reason.
       marked as expected-failing with a citation to F3
 - [ ] `npm test` passes from a clean clone with no `.env`
 - [ ] `app/.env.example` exists; `app/README.md` is project-specific
-- [ ] `POLICY_VERSION` bumped; drift guard in `npm run check`
+- [ ] `POLICY_VERSION` bumped; the drift guard runs **in CI** against an explicit base SHA
+      (not in `npm run check`, which stays local-only and has no base ref)
 - [ ] `ruff` and `mypy` run in CI
 - [ ] `garmin_login.py` removed
 
--- a/docs/plans/phase-0-instrumentation.md
+++ b/docs/plans/phase-0-instrumentation.md
@@ -8,6 +8,25 @@
 
 ---
 
+## Task board
+
+Status legend: `[ ]` not started · `[-]` in progress · `[x]` finished.
+Update the marker on the work-item heading **and** this table in the same commit.
+
+| Task | Status | Summary | Primary files |
+|---|:--:|---|---|
+| 0.1 | `[ ]` | Coaching invariants + aggregate bounds gate CI; snapshot becomes a non-blocking semantic diff | `app/scripts/simulate-scenarios.mjs`, `app/package.json`, `.github/workflows/ci.yml`, `docs/analysis/simulation-baseline.json` |
+| 0.2 | `[ ]` | Golden coaching-contract scenario with fixed dates; one assertion lands expected-failing | `app/src/engine/simulation/scenarios.ts`, `app/src/engine/goldenWeek.test.ts` |
+| 0.3 | `[ ]` | Lazy Firebase init so the suite runs from a clean clone | `app/src/firebase.ts`, `app/src/services/*.ts`, `app/.env.example`, `app/README.md` |
+| 0.4 | `[ ]` | `POLICY_VERSION` bump + CI drift guard using the CI base SHA | `app/src/engine/policy.ts`, `.github/workflows/ci.yml`, new guard script |
+| 0.5 | `[ ]` | `ruff` + `mypy` in `pyproject.toml` and CI | `pyproject.toml`, `.github/workflows/ci.yml` |
+| 0.6 | `[ ]` | Delete `garmin_login.py`; pin `uv`; add dependency audits | `garmin_login.py`, `Dockerfile`, `.github/workflows/ci.yml` |
+
+**No task here changes engine behaviour.** If a change to `app/src/engine/**` becomes
+necessary to complete a task, stop — it belongs in a later phase.
+
+---
+
 ## Goal
 
 Make it possible to see whether a planning change improved or degraded the recommended
@@ -30,7 +49,7 @@ was ever committed.
 
 ## Work items
 
-### 0.1 — Invariants are the CI gate; the snapshot is a diagnostic
+### `[ ]` 0.1 — Invariants are the CI gate; the snapshot is a diagnostic
 
 **This item was restructured after PR #5 review.** The original version proposed a
 full-output snapshot diff as the CI gate. That is the wrong contract here, and the
@@ -69,7 +88,7 @@ so it should be stable — verify rather than assume. If it is not stable, the s
 worthless as a diagnostic; fix the nondeterminism or drop the snapshot and keep only the
 invariants.
 
-### 0.2 — Golden coaching-contract scenario
+### `[ ]` 0.2 — Golden coaching-contract scenario
 
 The existing scenario assertions in `app/src/engine/scenarios.test.ts` check engine
 semantics (no constraint violations, no absurd streaks). They do not assert that the
@@ -98,7 +117,7 @@ F3. Land it as a documented failing assertion (`it.fails(...)`, with a comment c
 F3 and this plan) so Phase 3 has an unambiguous definition of done, rather than
 weakening the assertion to match current behaviour.
 
-### 0.3 — Make the frontend suite runnable from a clean clone
+### `[ ]` 0.3 — Make the frontend suite runnable from a clean clone
 
 `app/src/firebase.ts` calls `initializeApp` / `getFirestore` / `getAuth` at module scope,
 so importing any service transitively initialises Firebase. Without `VITE_FIREBASE_*`,
@@ -127,7 +146,7 @@ instructions: install, env, `npm run check`, `npm run dev`, `npm run test:rules`
 
 **Acceptance:** `git clone && cd app && npm ci && npm test` passes with no `.env` present.
 
-### 0.4 — Policy version guard
+### `[ ]` 0.4 — Policy version guard
 
 `POLICY_VERSION` (`app/src/engine/policy.ts`) is documented as "increment whenever a
 change can alter a persisted recommendation decision" and has never moved, including
@@ -156,7 +175,7 @@ through `HEAD`'s new ranking tie-break. A frozen string makes `replay.ts`'s
    This check belongs in CI rather than `npm run check`, since it needs the base ref;
    keep `npm run check` local-only.
 
-### 0.5 — Python lint and type checking
+### `[ ]` 0.5 — Python lint and type checking
 
 `AGENTS.md` requires type hints across all Python modules; nothing enforces it.
 
@@ -165,7 +184,7 @@ Add to `pyproject.toml` dev dependencies: `ruff`, `mypy`. Add `[tool.ruff]` and
 `src/garmin_sync` only). Add both to the `python-tests` CI job. Fix whatever the first
 run surfaces, or record explicit per-module ignores with a reason.
 
-### 0.6 — Small cleanups
+### `[ ]` 0.6 — Small cleanups
 
 * Delete `garmin_login.py` (repo root). It duplicates
   `scripts/bootstrap_garmin_tokens.py` and bypasses the ADR-0002 guard by injecting
--- a/docs/plans/phase-0-instrumentation.md
+++ b/docs/plans/phase-0-instrumentation.md
@@ -1,0 +1,199 @@
+# Phase 0 — Instrumentation & developer baseline
+
+* **Status:** Ready
+* **Depends on:** nothing
+* **Unlocks:** Phases 3, 4, 5 (none of them can be evaluated without this)
+* **Addresses:** F11, F14, F15, part of F10
+* **Rough effort:** 1–1.5 days
+
+---
+
+## Goal
+
+Make it possible to see whether a planning change improved or degraded the recommended
+week, and make the repository runnable from a clean clone. Nothing here changes engine
+behaviour.
+
+## Why this is first
+
+Every later phase replaces a tuned heuristic with a different one. Today there is no
+committed artifact that would show the difference: `npm run simulate:scenarios` produces
+a genuinely rich report — objective resolution counts, constraint violations, fatigue
+tier distribution, anchor placement, fragile-selection diagnostics — and writes it to
+`app/artifacts/simulation-reports/`, which is **gitignored** (`app/.gitignore:14`).
+
+The 0.21× event-modality suppression in F3 is the concrete case: it is visible in
+`modalityDistribution` and `objectiveResolution`, and nobody saw it because no baseline
+was ever committed.
+
+---
+
+## Work items
+
+### 0.1 — Invariants are the CI gate; the snapshot is a diagnostic
+
+**This item was restructured after PR #5 review.** The original version proposed a
+full-output snapshot diff as the CI gate. That is the wrong contract here, and the
+objection is decisive: Phases 3, 4 and 5 deliberately change objective credit,
+anti-stacking, fatigue semantics and eventually the whole sequence planner. A
+byte-exact gate over that would (i) freeze today's known-bad behaviour as the reference,
+and (ii) turn every intentional improvement into baseline churn, which trains reviewers
+to rubber-stamp the update. The gate must assert **properties that should hold across all
+those changes**, not equality with today's output.
+
+**The gate — invariant assertions (blocking).** See 0.2. These are the contract.
+
+**Aggregate bounds (blocking, deliberately loose).** A small set of range assertions over
+`runAllScenarios` output, wide enough that a correct planner change passes and only an
+implausible one trips:
+
+| Metric | Bound | Rationale |
+|---|---|---|
+| `constraintViolations` | exactly 0, all scenarios | already the harness's own contract |
+| rest/recovery day share | 5–40 % | a planner recommending no rest, or mostly rest, is broken |
+| longest same-template streak across chained weeks | ≤ 3 | catches a regression of the "tempo trap" class |
+| objectives generated but never resolved in any scenario | ≤ 1 | catches an F2-class credit failure |
+
+**The snapshot — diagnostic, non-blocking.** Still commit
+`docs/analysis/simulation-baseline.json` (with `commit` and `capturedAt` stripped — they
+come from `gitCommit()` and the clock in `simulate-scenarios.mjs`). Add
+`npm run simulate:diff` producing a **semantic** diff — per-scenario changes in category
+and modality distribution, objective resolution, and fatigue-tier day counts — and print
+it in the PR. Do **not** fail CI on it. Updating the baseline is a normal, expected part
+of a planner PR; the reviewer reads the semantic diff to decide whether the change is
+what was intended.
+
+**Determinism check before committing the snapshot:** run `npm run simulate:scenarios`
+twice and diff. `pickTemplate` is date-hash seeded and scenarios use fixed `startDate`s,
+so it should be stable — verify rather than assume. If it is not stable, the snapshot is
+worthless as a diagnostic; fix the nondeterminism or drop the snapshot and keep only the
+invariants.
+
+### 0.2 — Golden coaching-contract scenario
+
+The existing scenario assertions in `app/src/engine/scenarios.test.ts` check engine
+semantics (no constraint violations, no absurd streaks). They do not assert that the
+week is *coached well*. Add one scenario that does.
+
+Add to `app/src/engine/simulation/scenarios.ts` a scenario `cycling_a_event_build_week`:
+an A-priority `cycling_event` roughly 60 days out, `free_weights: true`,
+`indoor_bike: true`, `environment: 'either'`, weekday 60 / weekend 150 minutes, stable
+moderate readiness.
+
+Assert, in a new `app/src/engine/goldenWeek.test.ts`:
+
+| Invariant | Assertion |
+|---|---|
+| Key cycling quality is spaced | ≥ 48 h between any two picks whose `category` is `Hard Endurance`, `Moderate Endurance` or `Race-Specific Endurance` and `modality` is `Cycling` |
+| Anchors are protected | no `Lower-body Strength` / `Full-body Strength` with `systemicCost >= 0.5` on the day before or after a key cycling day |
+| Event modality is not penalised for frequency | the week contains **≥ 3** `Cycling` sessions |
+| Required objectives resolve | `threshold_quality` and `strength_maintenance` reach `completedExposures >= targetExposures` within the 7-day strip |
+| Rest exists | ≥ 1 `Rest` or `Mobility/Recovery` day |
+
+**Expect the third row to fail today.** That is the point — it is the executable form of
+F3. Land it as a documented failing assertion (`it.fails(...)`, with a comment citing
+F3 and this plan) so Phase 3 has an unambiguous definition of done, rather than
+weakening the assertion to match current behaviour.
+
+### 0.3 — Make the frontend suite runnable from a clean clone
+
+`app/src/firebase.ts` calls `initializeApp` / `getFirestore` / `getAuth` at module scope,
+so importing any service transitively initialises Firebase. Without `VITE_FIREBASE_*`,
+`trainingSettingsService.test.ts` fails to collect (verified). CI hides this by injecting
+dummy values; `predev` runs `npm run check`, so a fresh clone cannot start the dev server
+either.
+
+Convert to lazy accessors:
+
+```ts
+let _app: FirebaseApp | undefined;
+function app() { return (_app ??= initializeApp(firebaseConfig)); }
+export function getDb() { return getFirestore(app()); }
+export function getAuthInstance() { return getAuth(app()); }
+```
+
+Update the call sites (`services/*.ts`, `App.tsx`, `main.tsx`, `visual/*`) to call the
+accessor at use time rather than importing a module-scope constant. Keep `export const db`
+temporarily as a deprecated getter if the diff would otherwise be large; remove it in the
+same phase.
+
+Then add `app/.env.example` documenting the six `VITE_FIREBASE_*` keys, and replace
+`app/README.md` — currently the **unmodified Vite starter template** — with real setup
+instructions: install, env, `npm run check`, `npm run dev`, `npm run test:rules`
+(requires Java + emulator), `npm run simulate:scenarios`.
+
+**Acceptance:** `git clone && cd app && npm ci && npm test` passes with no `.env` present.
+
+### 0.4 — Policy version guard
+
+`POLICY_VERSION` (`app/src/engine/policy.ts`) is documented as "increment whenever a
+change can alter a persisted recommendation decision" and has never moved, including
+through `HEAD`'s new ranking tie-break. A frozen string makes `replay.ts`'s
+`policyMatchesCurrent` check meaningless.
+
+1. Bump it now to reflect the tie-break already shipped.
+2. Add a check to `npm run check` that fails when any of `rules.ts`, `optimizer.ts`,
+   `microcycle.ts`, `periodization.ts`, `fatigue.ts`, `planner.ts` or `dose.ts` differ
+   from the merge base without `policy.ts` also changing. A short script comparing
+   `git diff --name-only origin/main...HEAD` is sufficient; it does not need to be clever.
+
+### 0.5 — Python lint and type checking
+
+`AGENTS.md` requires type hints across all Python modules; nothing enforces it.
+
+Add to `pyproject.toml` dev dependencies: `ruff`, `mypy`. Add `[tool.ruff]` and
+`[tool.mypy]` sections (start permissive — `disallow_untyped_defs` on
+`src/garmin_sync` only). Add both to the `python-tests` CI job. Fix whatever the first
+run surfaces, or record explicit per-module ignores with a reason.
+
+### 0.6 — Small cleanups
+
+* Delete `garmin_login.py` (repo root). It duplicates
+  `scripts/bootstrap_garmin_tokens.py` and bypasses the ADR-0002 guard by injecting
+  `APP_USER_ID="bootstrap_user"`. Confirm the bootstrap script covers the same flow
+  first.
+* Add `npm audit --audit-level=high` and `uv run pip-audit` (or equivalent) to CI as
+  non-blocking informational steps.
+* **Pin `uv` in the Dockerfile.** `RUN pip install --no-cache-dir uv` is unpinned, so the
+  image build tracks whatever uv releases. The `readme = "README.md"` build failure fixed
+  in PR #5 surfaced with no repository change at all, which is exactly this class of
+  problem. Pin to a known-good version and bump deliberately.
+
+---
+
+## Acceptance criteria
+
+- [ ] invariant suite (0.2) + aggregate bounds run in CI and are blocking
+- [ ] `npm run simulate:diff` emits a semantic diff and is explicitly non-blocking
+- [ ] `docs/analysis/simulation-baseline.json` is committed and reproducible twice in a row
+- [ ] `goldenWeek.test.ts` exists; the event-modality-frequency assertion is present and
+      marked as expected-failing with a citation to F3
+- [ ] `npm test` passes from a clean clone with no `.env`
+- [ ] `app/.env.example` exists; `app/README.md` is project-specific
+- [ ] `POLICY_VERSION` bumped; drift guard in `npm run check`
+- [ ] `ruff` and `mypy` run in CI
+- [ ] `garmin_login.py` removed
+
+## Risks & rollback
+
+* **Invariants too strict.** An invariant that encodes today's accident rather than a
+  real coaching property will block a correct Phase 3-5 change. Each assertion in 0.2
+  must be defensible as coaching, not as "what the engine currently does" — if it cannot
+  be justified that way, it belongs in the non-blocking diagnostic instead.
+* **Snapshot nondeterminism.** If the report is not reproducible the semantic diff is
+  noise. Verify (0.1); if it fails, land the invariants and drop the snapshot rather than
+  shipping a diagnostic nobody can trust.
+* **Lazy Firebase touches many files.** Mechanical but wide. Land it as its own commit so
+  it can be reverted independently.
+* Everything here is additive or test-only. Rollback is per-commit.
+
+## Out of scope
+
+No engine behaviour changes. The golden test is expected to fail on one assertion; do not
+"fix" it here.
+
+## Docs to update
+
+* `docs/README.md` — link `docs/plans/`
+* `AGENTS.md` — add `simulate:diff`, `ruff`, `mypy` to the commands reference
+* `CLAUDE.md` — same
--- a/docs/plans/phase-0-instrumentation.md
+++ b/docs/plans/phase-0-instrumentation.md
@@ -39,7 +39,7 @@ Every later phase replaces a tuned heuristic with a different one. Today there i
 committed artifact that would show the difference: `npm run simulate:scenarios` produces
 a genuinely rich report — objective resolution counts, constraint violations, fatigue
 tier distribution, anchor placement, fragile-selection diagnostics — and writes it to
-`app/artifacts/simulation-reports/`, which is **gitignored** (`app/.gitignore:14`).
+`app/artifacts/simulation-reports/`, which is **gitignored** (`app/.gitignore`).
 
 The 0.21× event-modality suppression in F3 is the concrete case: it is visible in
 `modalityDistribution` and `objectiveResolution`, and nobody saw it because no baseline
--- a/docs/plans/phase-0-instrumentation.md
+++ b/docs/plans/phase-0-instrumentation.md
@@ -1,7 +1,7 @@
 # Phase 0 — Instrumentation & developer baseline
 
 * **Status:** Ready
-* **Depends on:** nothing
+* **Blocked by:** nothing
 * **Unlocks:** Phases 3, 4, 5 (none of them can be evaluated without this)
 * **Addresses:** F11, F14, F15, part of F10
 * **Rough effort:** 1–1.5 days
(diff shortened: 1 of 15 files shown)


Szczepanov (commented):
Strong review overall. I independently re-checked the main claims against base `2ea45cd`; F1, the zero-stimulus Garmin path in F2, F3's date-less modality history, F6's mutable audit record, the disconnected event-plan contract, and inert `intensityScale` all hold up. I would keep the central conclusion and the Plan-Intent/V2 direction.

I do think a few parts should be tightened before this becomes the repository's authoritative remediation document:

### 1. F2 practical-effect wording is too absolute
The sentence saying that for *"any athlete who does not answer the adherence prompt every single day: objectives never resolve"* overstates the mechanism. The verified bug is narrower and still serious: a **recognized Garmin-only completion without a matched structured/adherence stimulus earns zero objective credit** because the all-zero profile selects the vector path and blocks the fallback. Objectives can still resolve from other structured/followed exposures (or other fallback inputs), so missing one adherence response does not imply the whole ledger can never resolve.

Suggested wording: *"Recognized Garmin-only sessions without a matched structured/adherence stimulus earn no objective credit. If those are the athlete's primary completion source, the ledger can remain unresolved despite real training."*

This matters because the current wording makes the consequence sound broader than the evidence supports.

### 2. Phase-1 F2 fix should not intentionally rely on the keyword fallback
The proposed step says to omit an all-zero `stimulusProfile` *"so Garmin-only events fall through to the fallback path"* and then add `DEFAULT_STIMULUS_BY_MODALITY`. The second half is the right direction; the first should be described only as a guard against treating an unknown vector as known, **not as the desired credit path**.

The review itself correctly shows the keyword fallback is directionally wrong. For a recognized hard ride, `completedEventToExposure` builds a record like `"Cycling hard"`; if that reaches `updateMicrocycleProgress`, it can credit `threshold_quality` because it contains `hard` **and** `zone2_aerobic` merely because it contains `cycling`. That can replace today's false-negative bug with false-positive/double credit.

I would make the remediation explicit:
- zero stimulus means *unknown*, not a valid structured vector;
- recognized Garmin sessions should receive a coarse inferred stimulus vector with an evidence/confidence marker;
- keyword matching is legacy/last-resort compatibility only, not the intended Garmin path;
- add a regression test that a generic `Cycling hard` record does **not** automatically resolve both Z2 and threshold unless the inferred structured stimulus supports both.

### 3. Be careful making an exact simulation snapshot the Phase-0 CI gate
I agree completely that a golden coaching scenario belongs *before* planner changes. I would separate that from a byte-for-byte/deterministic recommendation baseline.

The next phases deliberately change objective credit, anti-stacking, fatigue semantics, and ultimately the sequence planner. A full-output snapshot gate risks freezing today's known-bad behavior and turning each intentional planner improvement into baseline churn. The stronger CI contract is the invariant layer you already describe: key-session spacing, no heavy lower-body interference, event-specific coverage, no third-cycling penalty, taper behavior, etc., plus bounded aggregate metrics where useful.

If a detailed simulation snapshot is retained, I would make it diagnostic or require an explicit reviewed baseline update with a semantic diff, rather than treating every recommendation change as a regression.

### 4. F12 identifies real model risks, but the replacement formula is not yet justified
The ordering invariant is a concrete correctness issue and should be fixed. The saturation and `max(external, internal)` behavior are also valid things to question. However, prescribing `1 - exp(-x)` plus a weighted external/internal combination is stronger than the evidence in this review supports.

In particular, internal response (HRV/RHR/soreness) can partly be a response to the same external work, so a weighted sum can double-count load unless calibrated. Also `1-exp(-x)` changes the state scale/meaning substantially; it is not automatically better just because it is monotonic.

I would reframe this remediation as: preserve an unsaturated latent external-load state (or otherwise retain depth), assert chronological ordering, then use the Phase-0 coaching/calibration harness to compare fusion functions before selecting one in an ADR. That is more consistent with F11's criticism of uncited tuned constants.

### 5. Minor wording nit in F1
`"An athlete with an achilles problem is offered heavy lower-body strength the moment they stop ticking the pain box"` should probably be `"can be offered..."`. Other readiness/fatigue/equipment gates may still suppress it; the verified defect is that the persistent/local injury gate no longer prevents it.

### CI note
At head `67f9603`, Python and Frontend/Firestore jobs are green, but **Docker Image Build is failing**, so the PR currently has a red CI pipeline. Since this PR changes only docs, I am not attributing that failure to the diff without the failed Docker log/root cause, but it should be resolved or explicitly identified as unrelated before merge.

With the points above tightened, I think this is a very useful architecture audit and a good basis for ADR-0010 and the V2 migration.

Szczepanov:
Thanks — all five points were right, and two of them were errors in the plans rather than wording. Addressed in `ec8a0fb`, alongside the new `docs/plans/` tree.

### 1. F2 wording — accepted, reworded

Adopted your framing almost verbatim. F2 now opens with a precise claim (a *recognised* Garmin-only completion with no matched structured/adherence stimulus earns zero credit) and the practical-effect paragraph now explicitly says objectives can still resolve from other sources and that a missed adherence response does not mean the ledger can never resolve. The consequence is scoped to "where recognised Garmin-only sessions are the athlete's primary completion source".

### 2. Keyword fallback as the credit path — accepted, this was a real bug in my plan

You're right, and this would have shipped a worse defect than the one it fixed. `"Cycling hard"` reaching `updateMicrocycleProgress` credits `threshold_quality` (contains `hard`) **and** `zone2_aerobic` (contains `cycling`) — false negative traded for double credit on one ride.

Phase 1.2 is rewritten around your four points: zero stimulus means *unknown*, not a valid vector; recognised sessions get a coarse inferred vector; keyword matching is last-resort legacy only. Added:
- your regression test — a generic `"Cycling hard"` record must not resolve both Z2 and threshold unless the inferred structured stimulus independently supports both;
- a `stimulusConfidence: 'exact' | 'inferred' | 'unknown'` field, so Phase 5's evidence hierarchy has a hook and we avoid a second migration;
- a note to check the proposed `moderate` row against `generateWeeklyObjectives`'s actual `targetStimulus` values, since that's exactly where the double-credit risk would reappear.

The ordering is now explicit: (b) is the fix, (a) is a correctness guard that rarely fires once (b) lands.

### 3. Simulation snapshot as CI gate — accepted, restructured

Agreed, and the "freezes today's known-bad behaviour" argument is the decisive one. Phase 0.1 is rewritten:

- **Blocking:** the golden-week coaching invariants, plus a small set of deliberately loose aggregate bounds (zero constraint violations; rest share 5–40%; longest same-template streak ≤ 3; unresolved-objective count ≤ 1).
- **Non-blocking:** the snapshot survives as `npm run simulate:diff`, emitting a semantic diff (per-scenario category/modality distribution, objective resolution, fatigue-tier counts) for the reviewer. Updating the baseline is a normal part of a planner PR.

Added a risk note that an invariant encoding "what the engine currently does" rather than a defensible coaching property will block correct Phase 3–5 work, and belongs in the diagnostic instead.

### 4. F12 replacement formula — withdrawn

Fair, and it was the same sin F11 criticises. Your double-counting point is the substantive one: internal response is partly a reaction to the same external work, so a weighted sum needs calibration to mean anything. Both the review and Phase 4.3 now split this:

- **(a) correctness, do now** — assert chronological ordering in `buildFatigueStateFromHistory` (throw, not comment).
- **(b) modelling, do not pre-decide** — retain an unsaturated latent external-load state so depth isn't discarded, then use the Phase 0 harness to *compare* fusion functions and record the comparison in an ADR.

The `1 - exp(-x)` prescription is explicitly marked withdrawn with the reasoning, rather than quietly deleted.

### 5. F1 wording — accepted

Now "**can be** offered", with the qualification that other gates may still suppress it and the verified defect is that the persistent local-injury gate isn't participating at all.

---

### CI: root-caused, fixed, and it isn't the diff

`pyproject.toml` declares `readme = "README.md"`; hatchling validates that during build; the Dockerfile copied only `pyproject.toml` and `uv.lock` before `uv sync`. Hence:

```text
OSError: Readme file does not exist: README.md
```

**Not attributable to this PR** — every recent `main` run is red, including base `2ea45cd`. Fixed anyway since it's small: copy `README.md`, and split dependency resolution (`--no-install-project`) from the project install so the dependency layer caches independently of `src/`. No Docker daemon in my environment, so this is reasoned rather than locally verified — CI is the check. Happy to split it into its own PR if you'd rather keep this docs-only.

The root cause is the unpinned `RUN pip install uv` — the build broke with no repository change at all. Recorded as a follow-up in Phase 0.6 rather than guessed at here.

---

### Also in this push: `docs/plans/`

Answering the "should we have `docs/plans`?" question — yes, on the grounds that ADRs are immutable decisions, analyses are dated audits, and plans are mutable and status-tracked. Six phase plans following the merged §7.5 ordering, each with preconditions, file-level work items, tests, acceptance checklists, risks/rollback, and out-of-scope. `docs/workout-library-expansion-plan.md` moved in as `0000-` to establish the convention, carrying a warning that its §1.2 line references are stale.

Phase 2 is deliberately `Draft` — it proposes `PlanDefinition` and the ADR-0010 domain model, and shouldn't go `Ready` without your agreement on the phase-vocabulary reconciliation (`EventPlanPhase` vs `PhaseWeights.phaseName`) and on whether `intensityScale` gets a consumer or gets deleted.

---

Szczepanov:
All plans promoted to **`Ready`** in `760258d`. Every open choice is decided, with the rejected alternative and the reason recorded next to it, plus a decision register in `docs/plans/README.md` so none of it has to be rediscovered across six documents.

| ID | Decision | Rejected alternative |
|---|---|---|
| **D-INJ** | Structured `InjuryConstraint` (Option B) | Consolidating onto guardrails (Option A) |
| **D-PHASE** | `EventPlanPhase` canonical; `PhaseWeights.phaseName` derived | Keeping the engine vocabulary canonical |
| **D-INT** | `intensityScale` gets a consumer | Deleting it |
| **D-TIER** | Build the next-day tier selector | Defaulting to `yellow` |
| **D-AXES** | Templates declare the axes; derivations deleted | Finding a citation for `*0.8` / `*0.7` |
| **D-BEAM** | Build *and measure* beam search | Pre-committing to shipping it |

The three where the reasoning is load-bearing rather than obvious:

**D-INT — `intensityScale` gets a consumer rather than being deleted.** The deletion case is that nothing reads it, so it's dead weight. The counter-case wins: taper is *defined* by volume and intensity moving in opposite directions, and `plannedDose` currently collapses both into one `volumeScale`-derived scalar. That's exactly why taper behaviour has to be reconstructed downstream from template `phaseEligibility` and ranking weights — the emergent-taper problem in Phase 5.7. Deleting `intensityScale` would make a correct taper harder to express, not simpler. It splits into `PlannedDose { volume, intensity }`, consumed in Phase 4.4. It stays written-and-unread until then, which is acceptable only because it now has a named, scheduled consumer — the state F17 objects to it lacking.

**D-PHASE — `EventPlanPhase` becomes canonical.** The deciding factor isn't that it's the coach-facing vocabulary, it's that it can express `travel`. A days-to-event model structurally cannot represent travel, because travel has nothing to do with event proximity. Mapping the other direction loses that. The fallback mapping is tabulated; `race` and `travel` are only ever set by an explicit `PlanDefinition`, since inferring them would be a guess.

**D-BEAM — approved to build and measure, not to ship regardless.** This is the one place I've deliberately *not* closed the question, because it isn't mine to close by argument. Whether sequence search beats the patched greedy loop is empirical, and the Phase 0 invariants are what answer it. If it doesn't measurably improve them, recording that and keeping the greedy loop is a successful increment. Phase 5 also gains an explicit 7-step increment order — fixed activities first (highest value per effort, zero dependency on the search work), sequence search last.

Two smaller things folded in: the injury region→restriction table and the Garmin stimulus inference table are now **implementation defaults** rather than pending reviews — deliberately biased toward over-restriction, on the basis that a false exclusion costs one session and a false permission costs a re-injury. The double-credit verification against `generateWeeklyObjectives`'s real `targetStimulus` values stays a required implementation step, not an optional check, since that's where the keyword matcher's failure mode would sneak back in.

The register also records **D-KWD** and **D-GATE**, which came out of your review rather than from me.

CI green on `ec8a0fb`; re-running for `760258d`.

---

docs/plans/phase-1-live-defects.md
```diff
@@ -0,0 +1,374 @@
...
+a correctness guard, not a routing mechanism.
+
+Values must be deliberately conservative — the aim is "materially better than treating
+real training as adaptation-neutral", not physiological precision. **Adopted starting
+values** for `Cycling`; the other modalities follow the same shape:
+
+| Intensity | `aerobicCapacity` | `thresholdDevelopment` | `surgeRepeatability` |
+|---|---|---|---|
+| easy | 0.75 | 0.15 | 0.0 |
+| moderate | 0.70 | 0.60 | 0.25 |
```
Szczepanov: **The adopted `moderate` values already violate the required verification immediately below.** With today's V1 math, `stimulusCoverage` for a single-axis objective is effectively the candidate value itself. `zone2_aerobic` targets `{ aerobicCapacity: 0.8 }`, so `0.70` gives coverage `0.70 >= 0.6`. `threshold_quality` targets `{ thresholdDevelopment: 0.9 }` and requires threshold stimulus `>= 0.6`, so `0.60` gives coverage exactly `0.60` and passes the qualification gate. Therefore this one `moderate` ride resolves **both** objectives—the exact false-positive case the plan says must not happen.

Please don't publish these as “adopted starting values” while simultaneously instructing implementation to tune them later. Either choose a table that satisfies the current credit function now, or better, state that the values are illustrative and make the first implementation task a table-driven test over every modality×intensity×objective combination before constants are selected. This is a good example of why Phase 1.2 should depend on the Phase-0/credit regression harness.

docs/plans/phase-1-live-defects.md
```diff
@@ -0,0 +1,374 @@
...
+    &&  request.resource.data.category       == resource.data.category
+    &&  request.resource.data.modality       == resource.data.modality
+    &&  request.resource.data.mode           == resource.data.mode
+    &&  request.resource.data.rationale      == resource.data.rationale
+    &&  request.resource.data.createdAt      == resource.data.createdAt;
+}
+```
+
+`recommendationAudit` needs care: `Home.tsx` writes the record once with the audit
+attached, and `handleAdjustSession` re-saves the same base recommendation with an added
```
Szczepanov, line 497: **Pinning the decision fields creates a new same-day recomputation problem unless recommendation lifecycle semantics change too.** `Home.tsx` recomputes today's recommendation from the latest check-in/Garmin/settings on dashboard load, puts that new recommendation in UI state, and then calls `saveRecommendation`. `saveRecommendation` currently overlays the newly computed template/mode/rationale onto the existing `{date}` document. If these fields become immutable after the first write, a legitimate later same-day recomputation that changes the recommendation will be shown in the UI but rejected by Firestore, leaving persisted audit/history describing a different recommendation than the user sees.

The plan needs to choose the lifecycle explicitly before tightening the rules: **(A)** first persisted recommendation for a date is authoritative and subsequent loads reuse/replay it, or **(B)** recommendations are append-only revisions (e.g. `{date}/revisions/{revision}` or revision IDs) and the latest revision is what UI/adherence refers to. Simply pinning fields in the existing one-document model is not enough. Also decide whether `prescription` is immutable decision evidence; the proposed `decisionFieldsUnchanged()` does not pin it.

docs/plans/phase-2-plan-intent-authority.md
```diff
@@ -0,0 +1,222 @@
...
+because travel has nothing to do with event proximity. The reverse mapping loses that.
+
+Mapping for the generic fallback:
+
+| `PhaseWeights.phaseName` | `EventPlanPhase` |
+|---|---|
+| Base, Build | `build` |
+| Specificity | `peak` |
+| Peak/Taper | `taper` |
+| Post-Event Recovery | `build` (at reduced volume) |
```
Szczepanov: **`Post-Event Recovery → build` is a lossy canonicalisation in exactly the layer intended to become authoritative.** Once `EventPlanPhase` drives objectives/workout eligibility, calling recovery `build` risks making build-phase requirements eligible during the period where the generic engine currently deliberately emits `Post-Event Recovery`. “At reduced volume” does not restore the missing semantics—build still means develop fitness; recovery means suppress/defer it.

I would add a canonical `recovery` phase (or a separate block/lifecycle state that has equivalent authority) instead of aliasing it to `build`. Then explicitly map `Post-Event Recovery → recovery` and declare which coverage families are legal there. Otherwise the migration removes information from the fallback model while claiming to make the new vocabulary canonical.

docs/plans/phase-2-plan-intent-authority.md
```diff
@@ -0,0 +1,222 @@
...
+`WeeklyObjective` already declares `requiredCredit`, `priority`, `windowStart` and
+`windowEnd` and uses none of them (F17). This is where they get used — objective windows
+become real dated ranges derived from `PlanBlock`, replacing the generic rolling 7-day
+counter (`MicrocycleState.weekStartDate`, whose name already misdescribes it: rename to
+`windowStartDate`).
+
+### Migration path
+
+1. Keep `SEPTEMBER_CYCLING_EVENT_SESSION_COVERAGE` as data; add a pure adapter
+   `planDefinitionFromCoverage(coverage, event)` producing a `PlanDefinition`.
```
Szczepanov: **`planDefinitionFromCoverage(coverage, event)` cannot actually produce the `PlanDefinition` shown above.** The coverage data knows `phases`, requirement tier and workout IDs; it does **not** contain `PlanBlock.startDate/endDate`, `volumeScale`, `intensityScale`, or travel dates. A `UserEvent` gives you the event date, but it still cannot tell you that travel is e.g. Aug 19–22. This is important because the whole reason for making the plan explicit is to represent states that days-to-event cannot infer.

Please split *coverage* from *schedule*: keep coverage as reusable phase→workout-family knowledge, and provide a separate dated block schedule/macrocycle input (or a concrete `PlanDefinition` for this event) that supplies build/travel/peak/taper/race/recovery dates and scales. An adapter can combine those two sources, but coverage+event alone is insufficient and would force hidden inference back into the layer meant to remove it.

docs/plans/phase-2-plan-intent-authority.md
```diff
@@ -0,0 +1,222 @@
...
+**D2 — `intensityScale` gets a consumer; it is not deleted.**
+The deletion argument is that nothing reads it. The counter-argument wins: taper is
+*defined* by volume and intensity moving in opposite directions — volume down, intensity
+preserved. `plannedDose` currently collapses both into one scalar derived from
+`volumeScale` alone, which is precisely why taper behaviour has to be reconstructed
+elsewhere from template `phaseEligibility` and ranking weights (F17, and the emergent-taper
+problem in Phase 5.7). Deleting `intensityScale` would make a correct taper harder to
+express, not simpler.
+
+The consumer, specified here and implemented in Phase 4.4:
```
Szczepanov: **Cross-document dependency is currently broken:** this says the `PlannedDose.intensity` consumer is implemented in Phase **4.4**, and Phase 5's increment table also says taper depends on “Phase 4.4's `PlannedDose.intensity`”. But Phase 4.4 is currently only **Dose-sensitive cost**; it never introduces `PlannedDose`, consumes `PlanBlock.intensityScale`, or defines intensity-class gating. Its acceptance criteria don't mention it either.

Please add an explicit work item/acceptance criterion in Phase 4 (or move the consumer to the appropriate phase) so D2 has a real owner. Otherwise F17 remains scheduled in prose but absent from the executable plan—the same class of disconnected commitment this review is criticizing.

Szczepanov (commented):
Second-pass review after the new implementation-plan commits. This is materially stronger than the version I reviewed at `67f9603`: the previous F2 wording/fallback issue, simulation-gate concern, F12 formula overreach and F1 wording have all been addressed well, and CI is now green at `760258d` (including Docker).

I re-checked the new plans against the base implementation rather than treating them as aspirational prose. There are still several executable contradictions worth fixing before these plans become the repository's implementation authority. I left inline comments on the highest-value ones.

Two cross-cutting points in addition to the inline comments:

1. **The plan status model is internally inconsistent.** `docs/plans/README.md` defines `Ready` as “preconditions listed are met”, then marks all six phases Ready while Phase 2 depends on Phase 1, Phase 3 on Phase 0 + ADR-0010, Phase 4 on Phase 0 + Phase 2, and Phase 5 on Phases 0–4. Either introduce `Approved/Blocked` (or redefine `Ready` as design-approved but potentially blocked), or only mark a plan Ready when its dependencies have landed. Right now the status table cannot be used operationally as documented.

2. **Phase 1.2 actually depends on Phase 0 despite the Phase-1 header saying it does not.** Phase 1.2 introduces new inferred-stimulus constants, explicitly says to re-run the Phase-0 invariant suite and semantic diff before merge, and its risks section says to ship behind that diff. That makes Phase 0 a hard prerequisite for 1.2, even if 1.1/F1 and 1.3/F6 can proceed independently. I would express dependency per work item rather than saying Phase 1 as a whole depends on nothing.

The central direction remains good: fix live safety/audit defects, make Plan Intent authoritative, move to one lexicographic ranking path, then finish objective-credit V2 and only afterwards evaluate sequence search. The issues below are mostly places where the now-detailed plan needs one more pass to become executable without rediscovering design questions during implementation.

docs/plans/phase-5-sequence-planning.md
```diff
@@ -0,0 +1,218 @@
...
+generation, template `phaseEligibility` and utility ranking. `event-plan.ts` already names
+the real roles — taper sharpening, pre-race openers, race-week strength, race day. Use
+them, so taper directly expresses *reduce volume, preserve useful intensity and
+event-specific freshness* rather than arriving as an emergent side effect.
+
+---
+
+## Acceptance criteria
+
+- [ ] beam search replaces the greedy projected loop; hard constraints reject before scoring
```
Szczepanov: **These acceptance criteria contradict the explicit decision at the top of this plan.** Earlier, D-BEAM correctly says beam search is approved to be *built and measured*, not necessarily shipped; if it does not outperform the greedy planner, keeping greedy is defined as a successful outcome. But this checklist can only pass if beam search replaces greedy and a full week is sequence-scored.

Make 5.1's acceptance conditional: implement a comparable sequence-search prototype, benchmark it with the invariant/semantic harness, and then either **(a)** promote it and satisfy the sequence-specific criteria, or **(b)** record the result/ADR and retain greedy. The current checklist silently turns an experiment back into a predetermined migration.

docs/plans/phase-0-instrumentation.md
```diff
@@ -0,0 +1,199 @@
...
+`POLICY_VERSION` (`app/src/engine/policy.ts`) is documented as "increment whenever a
+change can alter a persisted recommendation decision" and has never moved, including
+through `HEAD`'s new ranking tie-break. A frozen string makes `replay.ts`'s
+`policyMatchesCurrent` check meaningless.
+
+1. Bump it now to reflect the tie-break already shipped.
+2. Add a check to `npm run check` that fails when any of `rules.ts`, `optimizer.ts`,
+   `microcycle.ts`, `periodization.ts`, `fatigue.ts`, `planner.ts` or `dose.ts` differ
+   from the merge base without `policy.ts` also changing. A short script comparing
+   `git diff --name-only origin/main...HEAD` is sufficient; it does not need to be clever.
```
Szczepanov: **This guard as written is brittle in the current GitHub Actions checkout.** `.github/workflows/ci.yml` uses `actions/checkout@v4` with default shallow history (`fetch-depth: 1`). In a PR job, `git diff --name-only origin/main...HEAD` is therefore not guaranteed to have `origin/main`/the merge base available and can fail or compare the wrong range.

Please make the plan explicit about supplying the base: either set `fetch-depth: 0`, explicitly fetch the PR base ref, or have the script accept `github.event.pull_request.base.sha`/`GITHUB_SHA` (with a separate push-to-main behavior). A policy-version guard that depends on ambient remote refs will become an intermittent CI guard rather than a reliable one.

Szczepanov:
Synced with `main` and addressed both review rounds in `1f9b55c`.

## main sync

Merged `dc62f97`. One conflict, in `docs/architecture/recommendation-engine.md`: main added a **Verification & Audit Tooling** section to the old doc while this branch had rewritten the file. Resolved by keeping the rewrite, taking main's tooling section, and **dropping main's stale metric block** (`HRV drop > 10%`, `RHR > baseline + 3 bpm`, `sleep score < 65`) — that block is F13 itself and would have reintroduced the finding.

Also worth noting: main's `app/README.md` rewrite covers part of Phase 0.3 already.

## Your two cross-cutting points — both correct

**Status model.** Guilty as charged; I created that contradiction when I promoted everything to `Ready`. Split into two axes: `Status` answers *is the design agreed*, `Blocked by` answers *can work start*. New lifecycle adds `Approved` between `Draft` and `Ready`. Result: **only Phase 0 and Phase 1 items 1.1/1.3 are `Ready` today**; 2–5 are `Approved (blocked)`.

**Phase 1.2 depends on Phase 0.** Right — the plan said "depends on nothing" in the header while 1.2's own text required the Phase-0 diff before merge. Dependencies are now per work item, and 1.2 carries an explicit blocked-by banner.

## Your inline comments

**The `moderate` values — you're right, and I checked the arithmetic:**

```text
zone2_aerobic      target {aerobicCapacity: 0.8}      → coverage 0.70 ≥ 0.6  ✓ credits
threshold_quality  target {thresholdDevelopment: 0.9} → coverage 0.60 ≥ 0.6  ✓ credits
                   qualification min 0.6              → 0.60 ≥ 0.6           ✓ passes
```

One ride, both objectives — the exact false positive the plan forbade two paragraphs later. Publishing constants while deferring their validation is the F11 practice, and doing it inside the F2 fix was worse. **No constants are adopted now.** The table is explicitly illustrative shape only, and the first implementation task is a table-driven test over every `modality × intensity × objective` cell, with the constants chosen to satisfy it. This is now the concrete reason 1.2 is blocked on Phase 0.

**The lifecycle problem is the one I'm most glad you caught.** Naive field-pinning would reject a legitimate midday recomputation (session completed → `alreadyTrainedToday` → mode flips to `recover`) while the UI shows it — audit contradicting the user's screen. Added **D-LIFE**: append-only revisions, decision fields immutable *per revision*, `prescription` pinned as decision evidence. One honest limit recorded: Firestore rules cannot verify the prior revision was actually copied — that's a client batch obligation covered by an emulator test, not by the rules, and ADR-0010's amendment will say so rather than implying more.

**`Post-Event Recovery → build`** — lossy, agreed. Added **D-RECOV**: `EventPlanPhase` gains a canonical `recovery` member, with the ADR required to declare which coverage families are legal there.

**`planDefinitionFromCoverage`** — you're right that it cannot exist. Coverage has no block dates or scales, and no event date implies travel is Aug 19–22. Split into reusable *coverage* and an authored dated *block schedule*, combined by `buildPlanDefinition`. The block schedule is new authored input, and I've said plainly there's nowhere else for it to come from.

**`PlannedDose` had no owner** — correct, and exactly the disconnected-commitment class this review criticises. Added **Phase 4.5** with its own acceptance criteria.

**Phase 5 acceptance vs D-BEAM** — fixed; acceptance is now conditional, and recording a negative result satisfies it.

## CodeRabbit (17 findings, all applied)

Two were real correctness catches: **`ankle` was missing from `BodyRegion`**, so the change would have *removed* an existing running restriction the old regex covered; and **injuries had two sources of truth** — the legacy `string[]` is now deleted rather than kept in parallel, which matters because it's the safety layer.

Also applied: policy guard can't rely on `origin/main` under shallow PR checkout (`fetch-depth: 1` confirmed in `ci.yml`); the emulator rules suite can pass while skipped; the 48h rule restated in calendar-day terms since `date` is date-only; key-cycling predicate defined; objectives reference `blockIds` not `phases` (a phase can recur); `EventTiming` ordering validated; stimulus-rename inventory expanded to every reader; unordered history sorts at ingestion rather than throwing into an outage; `FixedActivity` gets a user-owned path plus rules tests; tissue state may only *tighten* an injury constraint.

I did not chase markdownlint MD013 line-length — it fires on pre-existing docs too and there's no markdownlint config in the repo. MD040, which was actually flagged, is clean.

---

docs/plans/phase-1-live-defects.md
```diff
@@ -0,0 +1,471 @@
...
+> intact under (B): it attaches to the date, resolving to the latest revision.
+>
+> **Shape:** keep `users/{uid}/daily_recommendations/{date}` as the current decision, add
+> a monotonic `revision: number`, and require that any update changing decision fields
+> increments it while writing the prior state to
+> `users/{uid}/daily_recommendations/{date}/revisions/{revision}` in the same batch.
+> Rules enforce: revisions are create-only (never update or delete); `revision` strictly
+> increases; decision fields may only change when `revision` increases.
+>
+> **Honest limit:** Firestore rules cannot verify the prior state was actually copied —
```
Szczepanov: **[P1] Enforce the archive write in the rules.** Firestore rules can validate related writes in the same batch/transaction with `getAfter()` / `existsAfter()`. The update rule can require the expected revision document to exist and compare its pinned decision/audit fields to `resource.data` (the pre-update current decision). Leaving this as only a client-side batch obligation lets a malicious client replace the current recommendation without preserving the prior audit record—the integrity property this phase is intended to establish. Please define the revision numbering/path precisely and add rules + emulator tests for a current-document update without its archive write, and for a mismatched archive.

Szczepanov (commented):
One data-integrity concern in the append-only recommendation revision plan.

Szczepanov:
Pushed `8862ce9`. Three things: your `getAfter()` correction, task status tracking, and thread cleanup.

## Your P1 — you're right, and my caveat was the dangerous part

I wrote that Firestore rules "cannot verify the prior state was actually copied — that is a client-side batch obligation." **That's false.** Rules can validate another document's post-write state in the same batch via `getAfter()` / `existsAfter()`.

The bad part wasn't the inaccuracy, it was that I dressed it up as an "honest limit" — which would have shipped the exact hole this work item exists to close, with a note explaining why the hole was acceptable. A client could have replaced the current recommendation without preserving the prior audit.

Now specified concretely:

- **Path/numbering:** `users/{userId}/daily_recommendations/{date}` holds the current decision with monotonic `revision` (first write `1`); archives live at `.../{date}/revisions/{revision}`, doc ID = the **prior** revision as a string.
- **`archivesPriorRevision()`** compares the archive field-for-field against `resource.data` (pre-update), including `prescription` and `recommendationAudit`. A missing, mismatched, or fabricated archive is rejected.
- The lookup is bound once with `let` — each distinct `getAfter()` is a document access, and rules cap those (10 single-doc / 20 multi-doc or batched). Comparing nine fields via nine `getAfter()` calls would have risked hitting the ceiling.
- **Carve-out:** adherence answers and `adjustment` additions don't change decision fields, so they require `revision` unchanged and **no** archive write. Making the common path expensive is how clients end up routing around the rule.
- `revisions/{revision}` is create-only; never update, never delete.

Emulator tests now cover: decision change without archive, mismatched archive, wrong revision number, non-incrementing revision, and mutation of an archived revision.

The residual limit, stated accurately this time: rules constrain the archive's content and existence, but can't compel a client to write a new decision at all — which isn't a gap.

## Task status tracking

Every work item carries `[ ]` / `[-]` / `[x]` on its heading, plus a **Task board** per plan mapping each task to status, summary, and primary files. The boards also carry sequencing that was previously buried in prose:

- **1.2 blocked on Phase 0** (1.1 and 1.3 are not)
- **2.1 gates the rest of Phase 2** — the domain objects are the ADR's output
- **4.5 before 4.4** — both touch the dose call path, doing 4.4 first means reworking it
- **Phase 5 ordered by increment, not section number** — 5.3 first, 5.1 last

`docs/plans/README.md` documents the markers and requires heading + board updated in the same commit. One nuance recorded: a task is `[x]` only when its *Done when* holds — and for **5.1**, a recorded negative result satisfies completion, since D-BEAM approves measuring beam search, not shipping it. Without that, the checkbox would quietly re-convert the experiment into a migration.

## Threads

**12 of 26 were still unresolved** — 7 of yours, 5 from CodeRabbit — all fixed in code but never marked. Resolved. All 26 are now closed.

## What I'd do next

1. **Split the Dockerfile fix to its own PR off `main`.** `main` is red without it and this is otherwise docs-only.
2. **Start Phase 0** — the only fully unblocked plan, and it gates 3–5 plus 1.2.
3. One thing I deliberately did *not* do: `docs/workout-library.md` claims the resolver checks `status === 'active'` while `prescription.ts` only checks `!manualOnly`. A `draft`/`deprecated` workout carrying `engineTemplateIds` would be prescribed. Nothing exploits it today (all 30 workouts are `active`), but it's a latent trap. It's a code fix, not a docs fix — say the word and I'll add it to Phase 1 as 1.4.

---

Szczepanov:
Correction: **the resolver `status` finding I raised three times is false.** No task 1.4 — there is nothing to fix. Pushed `9523c13`.

`workoutForTemplate` ([`prescription.ts:303`](https://github.com/Szczepanov/adaptive-training-recommender/blob/9523c13/app/src/workouts/prescription.ts#L303)):

```ts
.filter((workout) => workout.status === 'active' && !workout.manualOnly && workout.engineTemplateIds?.includes(templateId))
```

and the legacy fallback on line 307 also requires `status === 'active'`. Both gates exist. The resolver does exactly what `docs/workout-library.md` describes.

**Where I got it wrong.** The claim comes from `0000-workout-library-expansion.md` §1.3, written *before* that plan was implemented. It was fixed during the implementation, but the archived document still carried it as outstanding with a "fix in Phase 1" instruction. I read the archived plan, repeated the finding in two PR comments and in chat, and never checked it against the code — while the review this PR contains exists precisely to catch documents that disagree with the code. I trusted a doc over the source three times in a row.

**Fixed at the source**, so the next reader doesn't repeat it:

- §1.3's finding is struck through and marked resolved, with the verification recorded inline. Struck rather than deleted so the history stays readable.
- The reader warning at the top previously covered only stale *line numbers*. That was too narrow — it's what let me treat the *findings* as current. It now says the whole document is a historical record rather than a work list, and that any finding in it must be verified against the code before being acted on.

Worth noting the general shape, since `docs/plans/` will accumulate more of these: an implemented plan's pre-implementation findings read exactly like open work. The `Implemented` status alone wasn't enough of a signal — I saw it and still mined the body for findings. If this recurs, the stronger fix is to strip or collapse the "problems found" sections when a plan flips to `Implemented`, keeping only the outcome. Happy to do that as a convention in `docs/plans/README.md` if you want it.

---

Szczepanov:
Two rounds pushed since the last update: `7426ffe` (docs optimized for agent consumption) and `9e6fff9` (this review round).

## `7426ffe` — agent consumption

Three changes aimed at an agent arriving cold with no idea which document to believe.

**Routing table at the top of `docs/README.md`.** The `docs/` directories are not interchangeable — `adr/` is intended design, `architecture/` is current behaviour, `analysis/` is dated evidence, `plans/` is future work — and reading one as another is how an already-fixed defect got re-reported three times in this PR. The table states what each directory is authoritative for, what it must *not* be trusted for, precedence when two disagree (code > `architecture/` > `adr/`), and task-oriented entry points ("changing engine behaviour → read these three, in this order").

**91 line references became symbol references.** A six-sample spot check found three already pointing at the wrong code *on the day they were written*. Both this and the "an `Implemented` plan must not read like a work list" rule are now recorded in `docs/plans/README.md` under a heading naming the incident that produced each.

**`AGENTS.md`** gains a pointer to the routing table, and its command list is deduplicated — `test:rules`, `simulate:scenarios` and `replay:recommendation` were each listed twice with differing descriptions.

## `9e6fff9` — 12 of 14 review findings

Adopted, grouped by what was actually wrong:

**Contradictions between a plan's body and its acceptance criteria.** Phase 0 required the `POLICY_VERSION` drift guard in `npm run check` while the body correctly places it in CI, since it needs a base ref.

**Rules stated as intent without the rule to satisfy them.** Phase 1 said every `RUNNING_INJURY_PATTERN` token must map to at least the same restriction — unsatisfiable for `leg` and `run`, which have no `BodyRegion` equivalent and never will. Both now map at the restriction level, with `region` optional where the site is genuinely unknown, plus required migration tests: without those two cases the migration passes everything else while quietly unblocking running for an injured athlete. Phase 2's `blockIds: string[]` left empty-list and non-contiguous-block semantics undefined with no defensible default — now exactly one block per objective, and an objective spanning a travel week is two objectives, which is also the honest model since credit earned before a travel week does not carry across it. Phase 5's multi-event design named "one authority, multiple contributors" without a total order; it now has one, with the final tiebreak commented as a determinism backstop rather than a real criterion.

**Validation that admitted the bug it was written to prevent.** Phase 2's `EventTiming` ordering checks all pass for a confirmed event tapering two weeks early — that is what happens whenever the confirmation handler writes `confirmedDate` and forgets `planningDate`. `confirmedDate` present now requires `planningDate === confirmedDate`. Phase 5's `FixedActivity` had ownership rules and validation for two fields; ownership does nothing about the owner's *own* client writing an unbounded `expectedCost`, which is a denial-of-service on the athlete's own schedule. Full field table now, with the note that rules cannot iterate a list, so item types are enforced in `validation.ts` — stated rather than left to assumption.

**Terms with no input.** Phase 4's cost formula named three factors against a two-factor `DeliveredDose`. Split: duration and completion ship now, measured response waits for its input contract. Writing it otherwise reproduces `intensityScale` (F17) — a term assigned six times and read zero.

**One that touches a hard invariant.** Phase 3's `dayDiff` must come from `app/src/utils/localDate.ts`. Inline `new Date(a) - new Date(b)` reads the strings as UTC midnight, so across a Warsaw DST boundary the difference is 47 or 49 hours and integer division silently yields the wrong answer — on exactly two days a year, in the rule deciding whether an athlete gets back-to-back hard sessions.

**Also adopted:** merged the duplicate `PlanBlock` declaration; specified the legacy stimulus read boundary (canonical wins, conflicts log, empty is `INVALID` rather than a zero profile that scores zero coverage forever, and the two invented `*0.8`/`*0.7` derivations are excluded from the mapping because D-AXES deletes them); and fixed the `Ready (1.1, 1.3) / Approved (1.2)` status, which was not one of the lifecycle values that same file defines. Plan status is one value again with item readiness in its own column.

## Two skipped

**`phase-1-live-defects.md` — "bind archive creation to the revision change".** Already specified, in the text the comment quotes: the update rule requires `request.resource.data.revision == resource.data.revision + 1` **and** `archivesPriorRevision(userId, date)`, which compares the archive field-for-field against the pre-update document. The supporting analysis observed that `app/firestore.rules` doesn't implement this — correct, and it is the point. This is a plan for work not yet done.

**`docs/plans/README.md` — "block Phase 3 on the ADR-0012 artifact".** Phase 3 is already blocked on it and marked `Approved`, not `Ready`; ADR-0012 is listed in its `Blocked by` cell precisely because it does not exist. I've made the cell say so in words rather than leaving the reader to infer it from a missing file.

**`phase-1-live-defects.md` — "add the recognized Garmin no-signal case"** was neither adopted nor skipped but answered: the state is unreachable, because the table is a total `Record<CompletedModality, Record<CompletedTrainingIntensity, …>>` and TypeScript rejects it if a cell is missing. That totality is load-bearing — if the state *were* reachable, a known `"Cycling hard"` event would fall through to the keyword matcher and collect exactly the double credit D-KWD exists to prevent. The plan now says so, forbids widening to `Partial<Record<…>>` or adding a lookup fallback, and requires a totality test so a future modality added to the union without a stimulus row fails at test time rather than at credit time.

---

PR #8, opened 2026-08-08
Phase 2 — Plan intent is the planning authority (ADR-0012, PlanDefinition, EventTiming, Envelope Extraction)

Szczepanov:
### Phase 2 Implementation: Plan Intent is the Planning Authority

Addresses F16, F17, and F9.

#### Key Changes
1. **Task 2.1 — ADR-0012 Accepted**: Recorded Decision D1 (canonical \EventPlanPhase\ including \
ecovery\), Decision D2 (\intensityScale\ consumer \PlannedDose.intensity\), lexicographic priority model hierarchy, layer boundaries, and amended ADR-0007 / ADR-0004.
2. **Task 2.2 — \PlanDefinition\ Model**: Created \planSchedule.ts\ with \PlanDefinition\, \PlanBlock\, \PlanObjectiveDefinition\, \uildPlanDefinition\ validation. Updated \microcycle.ts\ to derive dated objective windows from \PlanDefinition\ when present. Renamed \MicrocycleState.weekStartDate\ to \windowStartDate\.
3. **Task 2.3 — Race-Date Uncertainty (\EventTiming\)**: Added \EventTiming\ and \alidateEventTiming\ with date ordering checks and invariant \confirmedDate\ present implies \planningDate === confirmedDate\. Updated \periodization.ts\ to anchor on \planningDate\.
4. **Task 2.4 — Extract \evaluateReadinessAndSafetyEnvelope\ (F9)**: Extracted envelope calculation in \
ules.ts\ and collapsed double template evaluation in \evaluateTrainingWithIntent\. Updated \
ecommendation-engine.md\.

#### Verification
- \
pm run check\: 100% passed (301 unit tests, typecheck, lint, workout validation).
- \
pm run simulate:scenarios\ & \simulate:diff\: 0 semantic diffs against committed baseline.
- Plan status updated in \docs/plans/phase-2-plan-intent-authority.md\.

### DIFF
--- a/app/src/engine/architecture.test.ts
+++ b/app/src/engine/architecture.test.ts
@@ -220,6 +220,33 @@ describe('Architecture & Phased Engine Integration', () => {
             // Blended toward the cycling demand vector, not the flat default base demand
             expect(phase.phase.targetDemandVector.thresholdPower).toBeGreaterThan(0.5);
         });
+
+        // ADR-0012 Task 2.3 (EventTiming) -- goalToUserEvent is the only real production
+        // constructor of UserEvent, so timing must survive that hop unchanged.
+        it('goalToUserEvent carries timing through onto UserEvent when present', () => {
+            const timing = { earliestDate: '2026-09-05', latestDate: '2026-09-20', planningDate: '2026-09-05' };
+            const event = goalToUserEvent({
+                ...baseGoal, targetDate: '2026-09-05', eventCategory: 'cycling_event', timing,
+            });
+            expect(event!.timing).toEqual(timing);
+        });
+
+        it('goalToUserEvent omits timing entirely when the goal has none', () => {
+            const event = goalToUserEvent({ ...baseGoal, targetDate: '2026-09-13', eventCategory: 'cycling_event' });
+            expect(event!.timing).toBeUndefined();
+        });
+
+        it('evaluatePeriodizationPhase anchors on an unconfirmed timing.planningDate rather than the fallback targetDate', () => {
+            // planningDate (earliest possible date) is materially closer than the fallback
+            // targetDate -- if periodization.ts ignored timing, daysToEvent would read 37,
+            // not 10, and the taper/build phase read below would be wrong.
+            const event = goalToUserEvent({
+                ...baseGoal, targetDate: '2026-09-13', eventCategory: 'cycling_event',
+                timing: { earliestDate: '2026-08-17', latestDate: '2026-09-13', planningDate: '2026-08-17' },
+            })!;
+            const phase = evaluatePeriodizationPhase([event], '2026-08-07');
+            expect(phase.daysToEvent).toBe(10);
+        });
     });
 
     describe('Phase 3: Microcycle Objectives', () => {
--- a/app/src/engine/microcycle.test.ts
+++ b/app/src/engine/microcycle.test.ts
@@ -7,7 +7,7 @@ import { evaluatePeriodizationPhase } from './periodization.ts';
 
 describe('updateMicrocycleProgress', () => {
   it('credits controlled field work toward surge repeatability', () => {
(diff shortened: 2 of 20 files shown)


Szczepanov:
## Review follow-up: production wiring for PlanDefinition & EventTiming

Pushed [1dd9356](https://github.com/Szczepanov/adaptive-training-recommender/commit/1dd9356) addressing the review findings below.

### What the review found

`PlanDefinition` (Task 2.2) and `EventTiming` (Task 2.3) were fully unit-tested in isolation but unreachable from any real code path:

- `resolveTrainingIntent` (`trainingIntent.ts`) called `buildMicrocycleState` without ever passing a `planDefinition` — `buildSeptemberCyclingEventPlan`/`buildPlanDefinition` were called nowhere outside `planSchedule.ts` itself and the new unit tests.
- `UserGoal` had no `timing` field, and `goalToUserEvent` (the only real production constructor of `UserEvent`) never set `UserEvent.timing` — so `periodization.ts`'s `event.timing?.planningDate` always fell through to `event.date` for every real user.

This meant ADR-0012's "`event-plan.ts` has at least one engine-path consumer" (addressing F16) and the phase-2 doc's acceptance criteria were true only in a synthetic/test sense, not in production — and `simulate:diff`'s 0-semantic-diffs result was actually confirming *nothing could reach the new code*, not that it was safely integrated.

### Fixes pushed

**Production wiring**
- `UserGoal.timing: EventTiming` added, validated by `validateGoal`, persisted/cleared by `goalService.ts` the same way as `eventCategory`/`eventPreset`/`eventLifecycle`, and carried unchanged through `goalToUserEvent` onto `UserEvent.timing`.
- `resolvePlanDefinitionForEvent` (`planSchedule.ts`) resolves `buildSeptemberCyclingEventPlan`'s plan for the one event it was actually authored for, and is wired into `resolveTrainingIntent`. Documented as intentionally narrow — there's no generic per-event plan-authoring mechanism or UI yet (`Goals.tsx` still only collects a single `targetDate`), so every other event continues through the generic `daysToEvent` fallback.

**Design gap fix**
- `generateWeeklyObjectives`'s plan-derived branch returned *every* block's objectives regardless of date (a build-phase athlete would see peak/taper objectives two months out as already unresolved). Added an `asOfDate` parameter, threaded through `buildMicrocycleState`, that scopes to the block actually active on that date.

**Also surfaced while verifying**
- `check-policy-drift.mjs` against the true `gemini-phase1` base (not `gemini-phase2`, which is this PR's own head) shows the original PR modified `rules.ts`/`microcycle.ts`/`periodization.ts` — all decision-affecting per that guard — without bumping `POLICY_VERSION`. Bumped it to `2026-08-plan-intent-authority-v1`.
- Removed the dead `maxExecutionDose`/`restrictedModalities` fields from `evaluateReadinessAndSafetyEnvelope`'s return value (never read by either caller).
- `evaluateTrainingWithIntent`'s empty-candidates fallback no longer recomputes the envelope a second time (passes the already-computed state into `evaluateTraining`).
- Added an exhaustiveness guard to the `ObjectiveKey` switch in `microcycle.ts`.

**Docs**
- ADR-0012 gets a new "Production Wiring Scope" section spelling out what's now generic (`EventTiming`) vs. intentionally narrow (`PlanDefinition`).
- Phase-2 status doc's acceptance criteria updated to describe the production path, not just the unit tests.

### Verification
- `npm run check` — typecheck, lint, 306 tests (up from 301; net new coverage for the wiring + scoping), workout validation — all pass.
- `npm run simulate:scenarios` + `simulate:diff` — 0 semantic diffs. Confirmed this is because the new wiring is narrowly scoped to one specific event/date and correctly doesn't touch any of the 11 existing simulated scenarios (checked the baseline snapshot is byte-identical before/after), not because the new code is unreachable.
- `check-policy-drift.mjs origin/gemini-phase1` — now passes.

Everything else in the original PR (ADR-0012's decisions, the `PlanDefinition`/`EventTiming` domain models themselves, the `evaluateReadinessAndSafetyEnvelope` extraction) was solid as-is — this follow-up is purely about closing the gap between "the domain model exists and is unit-tested" and "a real user's data can actually reach it."

PR #11, opened 2026-08-08
Phase 4 — Objective credit V2 and honest load

Szczepanov:
## What changed

Implements Phase 4 objective-credit V2 and honest delivered-load work on top of the current `main`, including both CodeRabbit review passes and the Phase 3 stack-sync fixes:

- promotes fractional, dose-sensitive objective credit to the live microcycle ledger;
- makes endurance/power objective credit respond to delivered duration while keeping strength-maintenance duration-neutral until set/load evidence exists;
- validates every supplied persisted canonical/legacy stimulus axis as a finite `0..1` number and rejects malformed profiles as `DataState.INVALID`;
- canonicalizes engine-owned template stimulus axes into the same `0..1` contract;
- makes the legacy keyword fallback contribute a conservative `0.5` credit to the same authoritative fractional ledger, independent of replay order;
- uses one compatibility exposure projection for live and forecast paths;
- gives an active authored `PlanBlock` ownership of `PlannedDose { volume, intensity }` in explicit mode, bounded to the persisted contract (`volume 0..1`, `intensity 0..1.2`);
- rejects invalid execution-dose inputs instead of normalizing `NaN`/out-of-contract values;
- uses V2 fractional credit for planner display/ledger and stores future allocation in `projectedCredit` rather than `completedCredit`;
- scales completed-session cost down by delivered duration/completion ratio, while capping duration scaling at the fully delivered intended dose;
- parses persisted stimulus once per exposure before objective fan-out;
- preserves unsaturated external-load depth while retaining `max()` external/internal fatigue fusion after a recorded harness comparison;
- keeps triathlon planning from collapsing to a single supported modality by treating missing rolling sport-modality coverage as a Level-4 architecture signal;
- explicitly treats the former `2026-08-single-ranking-path-v1` policy as audit-only rather than pretending the current build can replay it;
- updates ADR-0006, ADR-0014, the recommendation-engine architecture doc, the Phase 4 plan, and the docs index to match the implemented contracts.

## Review status

All inline review threads on PR #11 are resolved. The stack conflict with the subsequently merged Phase 3 lineage was also resolved by transplanting/merging Phase 4 onto current `main`; GitHub currently reports the PR as mergeable.

Key review regressions now have direct coverage, including authored travel/taper dose ownership, invalid dose rejection, delivered-duration credit/cost scaling, malformed stimulus rejection, replay-order-independent legacy fallback, projected-credit ownership, hard-session intensity gating, strength-axis scoring, deterministic out-of-order history replay, and historical-policy replay rejection.

## Cutover evidence amendment

The original Phase 4 plan required a live V1/V2 shadow iteration before cutover. Review correctly identified that the implementation had already cut over while that checkbox remained open.

ADR-0014 and the Phase 4 plan now explicitly amend that gate: the known-defective V1 model is not kept live in parallel. Instead, expected V1→V2 divergences are a deterministic regression matrix (`phase4ReviewFixes.test.ts`): sub-0.6 fractional credit, delivered-duration sensitivity, malformed persistence rejection, replay-order-independent keyword compatibility credit, preserved qualification gates, and forecast-only `projectedCredit`. Any new divergence must be explicitly reviewed. The broader Phase 0 scenario harness remains a release gate and is not waived.

## Validation

Latest GitHub Actions run on the PR head:

- ESLint: pass
- workout validation: pass (52 exercises, 36 workouts, 16 parameter-binding sets, 17 September-event session families)
- policy-drift check: pass
- frontend/unit/integration tests: **368 passed, 14 skipped**
- Firestore emulator rules: **14/14 passed**
- Python checks: pass
- Docker build/check: pass
- scenario regression tests: pass
- aggregate scenario bounds gate: **fails only on overall rest/recovery share: 44.3% vs allowed 5–40%**

## Release note / remaining blocker

The aggregate recovery-share failure is deliberately not hidden by retuning the bound or normalizing the generated baseline. It is now the only failing CI gate and remains an explicit Phase 4 release blocker.

The original fusion experiment at the earlier reviewed Phase-4 boundary produced 58.9% rest/recovery days for current `max(external, internal)` versus 70.4% for capped addition. The current reviewed PR has improved to **44.3%**, but still exceeds the 40% ceiling. The supported conclusion remains: do not replace `max()` with the tested capped-addition candidate, and do not call `max()` safe/calibrated merely because that candidate was worse.

## Summary by CodeRabbit

* **New Features**
  * Training plans now account for planned and delivered workout volume and intensity.
  * Objective progress reflects fractional credit based on stimulus, duration, and completed training.
  * Recommendations better balance recovery, modality coverage, intensity, and event-specific goals.
  * Fatigue estimates use delivered training load and replay activities in chronological order.
* **Bug Fixes**
  * Improved validation prevents invalid dose and stimulus data from affecting plans.
  * Historical recommendation audits now clearly identify policies that cannot be replayed.
* **Documentation**
  * Updated architecture and decision records to describe objective credit, load tracking, and planning behavior.

### DIFF
--- a/app/firestore.rules
+++ b/app/firestore.rules
@@ -33,9 +33,17 @@ service cloud.firestore {
         && (adherence.notes == null || adherence.notes is string);
     }
 
+    function hasValidDose(dose) {
+      return dose is map
+        && dose.keys().hasOnly(['volume', 'intensity'])
+        && dose.keys().hasAll(['volume', 'intensity'])
+        && dose.volume is number && dose.volume >= 0 && dose.volume <= 1
+        && dose.intensity is number && dose.intensity >= 0 && dose.intensity <= 1.2;
+    }
+
     function hasValidRecommendationAudit(audit) {
       return audit is map
-        && audit.keys().hasOnly(['policyVersion', 'evaluatedAt', 'decisionContextRevision', 'safetyStatus', 'history', 'envelope', 'candidateScores'])
+        && audit.keys().hasOnly(['policyVersion', 'evaluatedAt', 'decisionContextRevision', 'safetyStatus', 'history', 'envelope', 'plannedDose', 'executionDose', 'candidateScores'])
         && audit.keys().hasAll(['policyVersion', 'evaluatedAt', 'decisionContextRevision', 'safetyStatus', 'history', 'envelope', 'candidateScores'])
         && audit.policyVersion is string
         && audit.evaluatedAt is string
@@ -57,6 +65,8 @@ service cloud.firestore {
         && audit.envelope.keys().hasAll(['safetyRestrictedModalityCount', 'planMaxAllowableTier'])
         && audit.envelope.safetyRestrictedModalityCount is int && audit.envelope.safetyRestrictedModalityCount >= 0
         && audit.envelope.planMaxAllowableTier in ['Rest', 'Mobility', 'Easy', 'Moderate', 'Hard']
+        && (!('plannedDose' in audit) || hasValidDose(audit.plannedDose))
+        && (!('executionDose' in audit) || hasValidDose(audit.executionDose))
         && audit.candidateScores is list && audit.candidateScores.size() <= 64;
     }
 
--- a/app/src/engine/architecture.test.ts
+++ b/app/src/engine/architecture.test.ts
@@ -30,10 +30,20 @@ function testContext(overrides: Partial<UserContext['constraints']> = {}, traini
     };
 }
 
+const ZERO_CANONICAL_STIMULUS = {
+    aerobicEndurance: 0,
+    thresholdPower: 0,
+    vo2MaxPower: 0,
+    repeatedSurges: 0,
+    sprintPower: 0,
+    fatigueResistance: 0,
+    maxStrength: 0,
+    hypertrophy: 0,
+};
+
 describe('Architecture & Phased Engine Integration', () => {
     describe('Phase 1: Schedule & Availability', () => {
         it("uses the athlete's own weekday/weekend budget from TrainingSettings, not a fabricated day-of-week table", () => {
-            // Saturday (2026-08-08) -> weekend default; Monday (2026-08-10) -> weekday default.
             expect(resolveAvailability('2026-08-08', null, [], testContext()).maxTimeMinutes).toBe(120);
             expect(resolveAvailability('2026-08-10', null, [], testContext()).maxTimeMinutes).toBe(45);
         });
@@ -41,10 +51,9 @@ describe('Architecture & Phased Engine Integration', () => {
         it("caps (rather than blindly overrides) today's check-in time with the athlete's own profile limit", () => {
             const todayCheckin = {
                 readiness: 5, sleepQuality: 5, fatigue: 5, soreness: 5, stress: 5, motivation: 5,
-                timeAvailable: 200, // Claims 200 min available
+                timeAvailable: 200,
                 painFlag: false, alreadyTrainedToday: false, preferredModalityToday: null,
             };
-            // Friday (weekday) profile limit is 45 -- wins over the 200 min claim.
             const availability = resolveAvailability('2026-08-07', todayCheckin, [], testContext());
             expect(availability.maxTimeMinutes).toBe(45);
         });
@@ -52,17 +61,11 @@ describe('Architecture & Phased Engine Integration', () => {
         it('deducts fixed activity duration and reserves capacity for future scheduled activities', () => {
             const fixedActivities = [
                 {
-                    id: 'fixed_1',
-                    title: 'Evening Football',
-                    date: '2026-08-12',
-                    durationMin: 90,
-                    isCompleted: false,
-                    expectedCost: { systemic: 0.8 },
+                    id: 'fixed_1', title: 'Evening Football', date: '2026-08-12', durationMin: 90,
+                    isCompleted: false, expectedCost: { systemic: 0.8 },
                 },
             ];
-
             const availability = resolveAvailability('2026-08-12', null, fixedActivities, testContext());
-            // Wednesday weekday budget is 45 min, minus 90 min fixed activity -> floored at 0
             expect(availability.maxTimeMinutes).toBe(0);
             expect(availability.reservedCapacityCost).toBe(0.8);
         });
@@ -73,7 +76,6 @@ describe('Architecture & Phased Engine Integration', () => {
                 expect(resolveAvailability(date, null, [], noKit).availableEquipment).toEqual([]);
             }
             const withBike = testContext({ hasIndoorBike: true });
-            // Equipment access doesn't depend on which day of the week it is.
             expect(resolveAvailability('2026-08-10', null, [], withBike).availableEquipment).toContain('indoor_bike');
             expect(resolveAvailability('2026-08-11', null, [], withBike).availableEquipment).toContain('indoor_bike');
         });
@@ -83,42 +85,24 @@ describe('Architecture & Phased Engine Integration', () => {
         it('evaluates A-event tapers and prevents C-events from hijacking taper phase', () => {
             const events: UserEvent[] = [
                 {
-                    id: 'c_race',
-                    title: 'Practice 5K',
-                    date: '2026-08-10', // 3 days away
-                    priority: 'C',
-                    lifecycle: 'scheduled',
-                    category: 'running_race',
+                    id: 'c_race', title: 'Practice 5K', date: '2026-08-10', priority: 'C', lifecycle: 'scheduled', category: 'running_race',
                     demandProfile: { aerobicEndurance: 0.8, thresholdPower: 0.9, vo2MaxPower: 0.9, repeatedSurges: 0.5, sprintPower: 0.5, fatigueResistance: 0.5, neuromuscular: 0.5 },
                 },
                 {
-                    id: 'a_race',
-                    title: 'Championship Marathon',
-                    date: '2026-09-06', // 30 days away (Specificity Phase)
-                    priority: 'A',
-                    lifecycle: 'scheduled',
-                    category: 'running_race',
+                    id: 'a_race', title: 'Championship Marathon', date: '2026-09-06', priority: 'A', lifecycle: 'scheduled', category: 'running_race',
                     demandProfile: { aerobicEndurance: 1.0, thresholdPower: 0.8, vo2MaxPower: 0.5, repeatedSurges: 0.3, sprintPower: 0.2, fatigueResistance: 0.9, neuromuscular: 0.3 },
                 },
             ];
-
             const phase = evaluatePeriodizationPhase(events, '2026-08-07');
-            // Primary event is A-race (30 days out), C-race 3 days away does not trigger taper override
             expect(phase.phase.phaseName).toBe('Specificity');
             expect(phase.phase.taperActive).toBe(false);
         });
 
         it('keeps passed scheduled events stale, while completed/DNF events alone unlock recovery', () => {
             const event = (id: string, lifecycle: UserEvent['lifecycle']): UserEvent => ({
-                id,
-                title: id,
-                date: '2026-08-06',
-                priority: 'A',
-                lifecycle,
-                category: 'cycling_event',
+                id, title: id, date: '2026-08-06', priority: 'A', lifecycle, category: 'cycling_event',
                 demandProfile: { aerobicEndurance: 0.8, thresholdPower: 0.7, vo2MaxPower: 0.4, repeatedSurges: 0.5, sprintPower: 0.2, fatigueResistance: 0.8, neuromuscular: 0.3 },
             });
-
             const scheduled = evaluatePeriodizationPhase([event('awaiting-outcome', 'scheduled')], '2026-08-07');
             expect(scheduled.phase.phaseName).toBe('Base');
             expect(scheduled.focusEvent).toBeNull();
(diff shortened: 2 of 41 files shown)


Szczepanov (commented):
Second-pass Phase 4 review against current head `220e100` and current `gemini-phase3` base.

The overall direction is right: one fractional ledger, canonical stimulus vocabulary, delivered-load-aware cost, unsaturated external load, and a split volume/intensity plan dose are all better contracts than the previous state. I would still treat this PR as **not merge-ready** for the reasons below.

## Blocking / release-level findings

### 1. V2 has already cut over while the mandatory shadow-mode criterion is still open

`docs/plans/phase-4-objective-credit-v2.md` explicitly says: **“Cut over only when the divergence is explainable.”** The acceptance checklist still has the V1-vs-V2 shadow comparison unchecked, yet `microcycle.ts` now makes `completedCredit`/`requiredCredit` authoritative.

The 58.9% scenario number in the PR is a fatigue/fusion comparison, not the required credit-model divergence report. Before merge, either:

- run V1 and V2 side by side and report per-scenario objective generation/resolution deltas with an explanation for the differences, or
- explicitly amend the plan/ADR to waive/change this gate.

Leaving the gate open while shipping the cutover contradicts the repository’s own implementation contract.

### 2. Explicit `PlanDefinition` still does not own `PlannedDose`

In `app/src/engine/trainingIntent.ts`, `resolveTrainingIntent` correctly resolves `planDefinition` and uses it to build the active objective block, but then does:

```ts
const plannedDose = resolvePlannedDose(periodization.phase, microcycle.objectives, unresolvedObjectives);
```

That throws away the authored block’s `volumeScale` / `intensityScale` and falls back to generic days-to-event periodization.

This is materially wrong for the authored September plan in `planSchedule.ts`:

- travel block = `0.6 / 0.8`
- taper block = `0.5 / 1.0`

while generic periodization on those same dates can still be around `1.0 / 1.1` or a much higher taper volume.

The same mistake exists in `planner.ts` for projected days:

```ts
plannedDose: resolvePlannedDose(periodization.phase, microcycle.objectives, unresolved)
```

Fix this with one shared resolver: **active PlanBlock scales in explicit mode, generic `periodization.phase` only in generic mode**. The acceptance test currently checks only `taper.volume < build.volume`, which passes even when the exact authored `0.5` taper scale is ignored. Add exact-value tests for travel/taper block ownership.

### 3. Live objective credit is not actually delivered-dose-sensitive

`deriveObjectiveCredit` in `app/src/engine/stimulus.ts` computes:

```ts
const earnedCredit = Math.round(rawStimulusContribution * completionRatio * 100) / 100;
```

but ignores `DeliveredDose.plannedDurationMin` and `completedDurationMin` entirely.

This matters because the live reconciliation path **does** populate planned/completed duration, but does **not** populate an independent `completionRatio`; that ratio therefore defaults to `1.0`.

Result: a 15-minute and a 75-minute session with the same inferred stimulus earn the same objective credit. That is exactly the defect Phase 4 says it is fixing (“15 abbreviated minutes should not equal a 75-minute race-specific ride”).

Either consume duration in an objective-appropriate way now, or narrow the PR/ADR claim and do not mark dose-sensitive credit complete. Add a regression test with identical stimulus and different completed duration.

### 4. `readStimulusProfile` does not actually reject malformed persisted stimulus data

The PR claims malformed persisted profiles become `DataState.INVALID`, but the implementation only checks whether *some key exists*. Values are then cast with `as number` without validating type, finiteness, or range.

For example:

```ts
{ aerobicEndurance: 'bad' }
```

reaches `AVAILABLE`. Downstream this can produce `NaN` credit. That is particularly dangerous because:

- `NaN <= 0` is false, so the credit is not rejected;
- `Math.min(requiredCredit, completedCredit + NaN)` stores `NaN`;
- `NaN < requiredCredit` is false, so `getUnresolvedObjectives` can make that objective disappear from the unresolved set.

Please validate every supplied canonical/legacy axis as a finite number in the documented stimulus range, define the partial-canonical policy explicitly, and add malformed-string / `NaN` / out-of-range tests. Null/empty coverage alone is insufficient for a persisted-data boundary.

### 5. Legacy fallback + fractional ledger is order-dependent

`updateMicrocycleProgress` still increments only `completedExposures`. `getUnresolvedObjectives` now does:

```ts
const completedCredit = objective.completedCredit ?? objective.completedExposures;
```

After any structured V2 event sets `completedCredit`, a later keyword-fallback completion changes only `completedExposures` and is therefore ignored by the authoritative ledger.

The reverse order behaves differently: if the legacy fallback lands first, a later structured credit seeds from `completedExposures` and can treat the objective as already resolved.

So identical evidence can produce different resolution depending on chronology/type ordering. If the keyword path remains a supported last-resort compatibility path, it must also update the authoritative fractional ledger using an explicitly documented conservative compatibility credit. Add structured→legacy and legacy→structured regression tests.

## Cross-cutting consistency issues

### Planner display/ledger still uses V1 semantics

`planner.ts` still determines `objectiveCredits` / `addressesObjectives` via:

```ts
stimulusCoverage(...) >= STIMULUS_CREDIT_COVERAGE_THRESHOLD
&& qualifiesForObjective(...)
```

while `applyPick` advances the microcycle through V2 `deriveObjectiveCredit`, which no longer uses that 0.6 coverage gate and can award partial credit.

The comment saying these paths “mirror exactly” is therefore no longer true. The user-visible/planning credit ledger can say “no credit” while the authoritative V2 ledger applied fractional credit. Use the authoritative credit function once and carry the earned fractional amount into the planner ledger.

### `projectedCredit` is declared but still has no owner

`WeeklyObjective.projectedCredit` is introduced specifically for future projection, but the week-ahead planner still mutates `completedCredit` for tomorrow/projected picks. The state is ephemeral, so this is not persistence corruption, but it defeats the model distinction and makes returned `microcycleObjectives` semantically ambiguous. Either wire `projectedCredit` into planning or remove/defer the field until there is a real consumer.

### Fusion evidence supports “retain max for now,” not “safe”

The comparison supports keeping `max()` as the status quo versus the tested capped-addition alternative. It does **not** establish `max()` as a safe model: both trajectories violate the aggregate recovery-share gate (58.9% vs 70.4%). I would phrase the conclusion as “no evidence to replace `max()` with this candidate” rather than “least-conservative tested safe option.”

## Validation state

At the time of this review there is no GitHub Actions PR workflow run attached to head `220e100`, so I cannot independently confirm the stated `npm run check` result from CI yet.

Overall: the Phase-4 architecture is moving in the right direction, but the current implementation is still only partially honoring the contracts it now documents. I would fix items 1–5 before merge, then re-run the scenario harness and the actual V1-vs-V2 credit shadow comparison before treating the cutover as complete.

Szczepanov:
## Root-caused and fixed the aggregate rest/recovery release gate

Pushed [a3d489a](https://github.com/Szczepanov/adaptive-training-recommender/commit/a3d489a).

### tl;dr

`PROJECTED_FATIGUE_RECOVER_THRESHOLD` in `planner.ts` had an internally-inconsistent derivation: its comment correctly identified 48h as the slowest modeled fatigue half-life, but then computed the actual constant using a *36h* worked example (~0.63) instead of the 48h one (~0.707) — and since gating takes the **max across all six dimensions**, the 48h group (impactTissue, lowerBody) is what actually decides the ceiling. The gate was firing on ordinary single-session fatigue, not genuinely saturated back-to-back load. Fixed by recalibrating to the correct half-life group. Aggregate rest/recovery share: **44.3% → 32.4%**, inside the [5%, 40%] bound, without touching the bound or the `max()` fusion formula.

### Why not the bound, why not the fusion formula

Your own boundary analysis in `docs/plans/phase-4-objective-credit-v2.md` already closed off both tempting shortcuts before I started, with real evidence:
- Phase 0 on `main` passed this exact gate at 34.8% — proof the bound itself is achievable.
- `max()` vs. capped-addition fusion was already A/B tested (58.9% vs. 70.4%) — `max()` wins, so that's not it either.

That left the actual threshold mechanism between those two ruled-out options unexamined.

### The bug

The comment on `PROJECTED_FATIGUE_RECOVER_THRESHOLD = 0.625` reasoned:

> "the slowest modeled half-life is 48h... 0.625 sits just below the one-day residual of a saturated **36h** dimension (~0.63)"

That's using the wrong half-life group for its own stated premise. The correct one-day residual for a **48h** dimension (the one that's actually slowest, and the one `maxFatigueDimension` will pick up whenever it's elevated) is `1.0 × 0.5^(24/48) ≈ 0.707`, not `0.63`.

I verified this empirically with temporary day-by-day instrumentation on `no_event_base_phase`: a single moderate run/ride (impactTissue cost ~0.5–0.7) plus one easy day was enough to push `impactTissue`/`lowerBody` to ~0.70 and keep it pinned there for most of a simulated month — additive per-session accumulation against the slow 48h decay rarely let it clear 0.625 again. That's exactly why `fatigueTierDayCounts` showed `modify: 0` for every scenario I checked: fatigue was oscillating between "fine" and "hard lockout" with no real middle ground, because the gate was tripping on ordinary training, not extreme load.

### Calibration

I tried `0.70` (exactly at the correct 48h boundary) first — it got the aggregate to 27.2%, but occasionally left a whole scenario-week with **zero** rest/recovery days, breaking the existing "at least one rest day per week" invariant (`goldenWeek.test.ts`, `scenarios.test.ts`). Overcorrected. Settled on **0.65** — between the two half-life groups' true residuals, with a real margin below 0.707 rather than the old value's razor-thin (and miscalibrated) gap. Every scenario keeps ≥1 rest/recovery day, and the aggregate lands at 32.4% (93/287).

### Also fixed: two masked, real compile errors

`npm run typecheck` (`tsc -b`) currently fails on this exact PR head:
- `microcycle.test.ts` / `phase3Review.test.ts` still use the legacy `WorkoutStimulusProfile` field names (`aerobicCapacity`/`thresholdDevelopment`/`surgeRepeatability`/`mobilityRecovery`) this PR's own canonical-only stimulus contract removed.
- `completedTraining.ts` has a real possibly-undefined access inside a type-predicate filter (the predicate's declared return type doesn't narrow the parameter's type *inside* the function body — that's TS's actual behavior, not a false positive).

These are currently invisible because `simulate:scenarios` (a real, non-`continue-on-error` step) runs and fails *before* the `build` step (which runs `npm run check`, including `typecheck`) in `.github/workflows/ci.yml`'s job order — so GitHub Actions never reaches the already-broken `build` step in the same run. Fixing the aggregate gate alone would have just moved the red X to a newly-reachable `build` failure. Fixed both.

### Docs

Updated the Phase 4 plan's boundary-analysis table and status line, and ADR-0014's two "still unresolved" references, to record the root cause and the fixed 32.4% figure.

### Verification
- `npm run check` — typecheck, lint, **368 tests**, workout validation — all pass.
- `npm run simulate:scenarios` — exit 0, aggregate 32.4%, every individual scenario has ≥1 rest/recovery day.
- `npm run simulate:diff` — baseline regenerated, diff clean.
- `check-policy-drift.mjs origin/main` — passes.

Szczepanov:
## Addressed both CodeRabbit reviews

Pushed [b9830e7](https://github.com/Szczepanov/adaptive-training-recommender/commit/b9830e7).

### Review reviewing the pre-existing PR code (10 items) — already resolved, no action needed

Checked each of the 10 items against current `HEAD` before touching anything. All 10 are already fixed by commits already on this branch (`23213be`..`b1d4280`), which predate my previous fix commit ([a3d489a](https://github.com/Szczepanov/adaptive-training-recommender/commit/a3d489a)) — confirmed both via `git log` and via CodeRabbit's own "✅ Addressed in commits …" marker on every one of those inline threads. Nothing left to do there.

### Review reviewing my own fix commit (2 items) — both fixed

**1. `calculateStimulusBenefit` didn't enforce `qualification.minimumStimulus` and omitted `vo2MaxPower`.**

Real bug: the ranking benefit function and the actual objective-credit ledger (`deriveObjectiveCreditFromProfile` in `stimulus.ts`) disagreed about which candidates could resolve which objectives. A candidate below an objective's `minimumStimulus` floor still got full ranking credit as if it could resolve it, and `vo2_max` objectives had no matching scoring term at all so no candidate could rank higher for carrying `vo2MaxPower` stimulus.

Fixed by reusing the existing shared `qualifiesForObjective` (`microcycle.ts`) — the same gate the credit ledger already uses — instead of re-implementing only half of it inline, and adding the missing `vo2MaxPower` scoring term. Added two regression tests covering both behaviors.

Fixing a real ranking bug changes rankings, so I traced (not reverted) three resulting test failures:
- One was my own new test's arithmetic mistake — fixed the expected value.
- `trainingIntentAcceptance.test.ts` expected a 37-day-out road race day-0 pick to land on Race-Specific Endurance; that category's only day-0-eligible template was winning purely off illegitimate credit toward `threshold_quality` (its `thresholdPower` is 0.4, below the 0.6 minimum). Hard Endurance genuinely qualifies for two objectives and wins post-fix on real merit — updated the assertion and documented why.
- `scenarios.test.ts`'s anchor-hit test: `cycling_criterium_A`'s exact-anchor-day hit rate went from partial to 0/4 while weekly rolling-window fulfillment went from partial to 4/4 — same root cause. Rewrote the assertions to describe the new, fully-deterministic split instead of the old partial-hit numbers.
- Also found and fixed, while investigating, an unrelated floating-point comparison bug in `scenarios.test.ts`'s `objectiveCreditTotal` helper (summed already-rounded credits without rounding the sum, so two mathematically-equal totals landed on opposite sides of a `<=` purely from IEEE754 summation order) — newly exposed by the ranking change, not caused by it.

**2. Self-contradictory doc text.**

`docs/plans/phase-4-objective-credit-v2.md` still said the aggregate rest/recovery failure "remains a release blocker," directly contradicting the very next paragraph (added in my previous commit) documenting the fix, in a plan whose status is `Implemented`. Rewritten as historical pre-fix context.

### Verification
- `npm run check` — typecheck, lint, **370 tests** (2 new), workout validation — all pass.
- `npm run simulate:scenarios` — exit 0, aggregate rest/recovery **30.0%** (86/287), inside [5%, 40%]; every scenario keeps ≥1 rest/recovery day.
- `npm run simulate:diff` — clean against the regenerated baseline.
- `check-policy-drift.mjs origin/main` — passes.

PR #17, opened 2026-08-09
Audit Phase 0–5 completion and start Phase 6 assurance

Szczepanov:
## Summary

- reconcile Phase 0–5 plan/task status with shipped code
- add Phase 6 correctness/coverage assurance and reviewed simulation-baseline controls
- align taper policy and aerobic-volume coverage with the active macrocycle
- add secure, explicit authored travel blocks to live planning

## Validation

Current head `f6364b8` — GitHub Actions run [31362918754](https://github.com/Szczepanov/adaptive-training-recommender/actions/runs/31362918754): **success**.

- frontend validation: **553 passed**, 32 emulator tests skipped in the normal unit invocation
- Firestore emulator suite: **32 passed**
- workout catalog validation: **52 exercises, 36 workouts, 16 parameter-binding sets, 18 September-event session families**
- `simulate:scenarios`: generated report only; committed baseline unchanged
- policy-drift guard: passed for decision-affecting changes
- production frontend build, Python tests, and Docker build: passed

## Remaining semantic acceptance gate

The committed simulation baseline has intentionally **not** been updated. `simulate:diff` still reports a materially higher rest/recovery share in several cycling scenarios (including stressed criterium), so the recovery-policy distribution requires explicit coaching review before baseline acceptance.


docs/analysis/2026-08-09-phase-0-5-completion-review.md
```diff
@@ -0,0 +1,62 @@
+# Phase 0–5 completion review — 2026-08-09
+
+## Scope and verdict
+
+This follow-up verifies the implementation plans derived from
+[`2026-08-08-architecture-review.md`](./2026-08-08-architecture-review.md) against
+PR #16's code, with the audit work on `codex/phase-0-5-completion-audit`. The earlier
+analysis remains a point-in-time record and is not edited.
+
+**All Phase 0–5 work items are implemented.** The original review is fully closed for
```
Szczepanov: This verdict is stronger than the current implementation supports. `generateWeekAheadPlan` in `planner.ts` explicitly documents a known Phase 5.6 gap: the multi-event objective set is seeded at `todayDate` and is **not recomputed when the taper authority or a contributor crosses its window inside the 7-day horizon**. Separately, `schedule.ts` explicitly says `FixedActivity.fixed` is “not yet consumed”; the activity's `environment`/`equipment` are persisted but do not alter `resolveAvailability`, and `reservedCapacityCost` is calculated but not consumed by ranking/fatigue. Those are Phase-5-scope behaviours, not F11/F12/F15 calibration work. Either narrow this claim and keep the affected Phase 5 acceptance items open, or explicitly move these residuals into Phase 6 before saying *all* Phase 0–5 work items are implemented.

docs/plans/phase-6-evidence-and-operational-assurance.md
```diff
@@ -0,0 +1,291 @@
...
+fixed synthetic inputs, and is classified as a blocking contract or an observational
+diagnostic.
+
+### 6.2a — Multi-event taper conflict
+
+Add `multi_event_taper_conflict`: an A-priority cycling event inside taper and a nearby
+scheduled B-priority event with overlapping demand.
+
+* Contract: `evaluatePeriodizationPhase` reports one taper authority.
+* Contract: `resolveMultiEventObjectives` includes eligible contributor demand, takes the
```
Szczepanov: The proposed static `multi_event_taper_conflict` case will not cover the known production gap in `planner.ts`: contributor objectives are resolved once at the seed date and carried unchanged through the horizon. Please add a scenario where the authority enters taper (or a contributor enters/leaves the contribution window) on day 2–6, and make the task explicitly fix/re-resolve objective admissibility across that transition. Otherwise Phase 6 can pass this contract while the documented Phase 5.6 gap remains untouched.

docs/plans/phase-6-evidence-and-operational-assurance.md
```diff
@@ -0,0 +1,291 @@
...
+  maximum rather than sum on collisions, and records a taper-incompatible quality drop.
+
+### 6.2b — Travel and fixed-activity week
+
+Add `travel_fixed_activity_week`: a fixed travel activity with an
+`availabilityOverride`, constrained environment, and unavailable equipment.
+
+* Contract: `resolveAvailability` applies the restrictive budget before selection.
+* Contract: selected templates fit time, environment, and equipment.
```
Szczepanov: This is currently more than a scenario-coverage task. `resolveAvailability` does not consume a `FixedActivity`'s `environment` or `equipment`; it derives equipment only from the athlete's standing constraints. Likewise the fixed activity's `expectedCost` becomes `reservedCapacityCost`, but that value is not used by `rankCandidates`/fatigue, so the stated next-day-fatigue observation cannot reflect the booked activity's planned load. Please make 6.2b include the missing planner wiring (or explicitly mark these as expected-failing defects to be implemented elsewhere), rather than describing them only as contracts to observe.

docs/plans/phase-6-evidence-and-operational-assurance.md
```diff
@@ -0,0 +1,291 @@
+# Phase 6 — Evidence-driven calibration and operational assurance
+
+* **Status:** Draft
+* **Blocked by:** approval of the measurement contract for Tasks 6.1–6.3; production
+  Firebase project identity and deployment owner for Task 6.4
+* **Unlocks:** evidence-backed changes to fatigue fusion and tuned decision thresholds
+* **Addresses:** remaining portions of F11, F12, and F15 in
```
Szczepanov: If 6.2 is intended to own the mid-horizon multi-event and fixed-activity integration gaps above, this `Addresses` line is incomplete: those are carryovers from Phase 5, not remaining portions of F11/F12/F15. Add explicit Phase-5 carryover ownership here/task-board entries, or keep Phase 5 from being declared fully implemented in the completion audit.

docs/plans/README.md
```diff
@@ -50,8 +50,10 @@ These implement the way forward in
 [`docs/analysis/2026-08-08-architecture-review.md`](../analysis/2026-08-08-architecture-review.md)
 §7.5. Finding IDs (`F1`, `F16`, …) refer to that document.
```
Szczepanov, line 52: Now that the table contains Phase 6 as a Draft, `All six phases are implemented` is ambiguous/misleading. This means **Phases 0–5**, while there are seven numbered phase rows (0–6). Please say `Phases 0–5 are implemented` here and in the later `All six phases are complete` sentence so the status summary cannot be read as including Phase 6.

Szczepanov (commented):
Reviewed the completion audit against the merged Phase 5 implementation, not just the changed prose. CI run #171 is green. The comments below are documentation/closure correctness issues: the current code still contains explicit Phase 5 gaps that the audit now declares fully implemented, and the Draft Phase 6 plan does not yet own all of them.

docs/analysis/2026-08-09-phase-0-5-completion-review.md
```diff
@@ -0,0 +1,62 @@
...
+F1–F10 and F13–F17. F11, F12, and F15 are only partially closed: the harness and key
+correctness work landed, but the remaining calibration, fatigue-fusion, and operational
+controls are not represented as finished work.
```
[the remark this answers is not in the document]
Szczepanov: Fixed in 1119f3f: excluded F15 from the fully-closed finding range (now `F1–F10, F13–F14, F16–F17`), and `docs/plans/README.md` now says "Phases 0–5 are implemented" / "Phases 0–5 are complete; Phase 6 remains Draft" in both places you flagged.

docs/plans/phase-1-live-defects.md
```diff
@@ -1,13 +1,15 @@
...
-  **1.2 (F2 objective credit)** requires Phase 0's invariant + credit-regression harness
+* **Status:** Implemented
+* **Blocked by:** nothing outstanding
 * **Unlocks:** a correct foundation for the Phase 2–5 cutover
 * **Addresses:** F1, F2, F6
 * **Rough effort:** 3–4 days
 
+> **Historical implementation record.** All work items and acceptance criteria below
+> were verified as delivered on 2026-08-09. They are retained for rationale and are
+> not instructions for new work.
```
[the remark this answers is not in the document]
Szczepanov: Fixed in 1119f3f: retitled the Pre-implementation-state / Work items / Tests / Fix / Current-state sections in `phase-1-live-defects.md` and `phase-2-plan-intent-authority.md` as historical/as-delivered (e.g. "Work items (historical checklist — completed)", "Fix (as delivered)"), and rewrote the imperative "Generalise `event-plan.ts`" sentence in 2.2 as a completed outcome.

docs/analysis/2026-08-09-phase-0-5-completion-review.md
Szczepanov: Agreed — narrowed the verdict instead of reopening Phase 5. In 1119f3f, `docs/analysis/2026-08-09-phase-0-5-completion-review.md` now explicitly calls out both gaps you cited (planner.ts's `todayDate`-seeded multi-event objective set not being re-resolved mid-horizon; schedule.ts's unconsumed `FixedActivity.environment`/`equipment`/`reservedCapacityCost`) as Phase-5 residuals that are *not* folded into F11/F12/F15, added them to the finding-reconciliation table and the Remaining-work list, and handed explicit ownership to Phase 6 tasks 6.2a/6.2b rather than claiming full closure.

docs/plans/phase-6-evidence-and-operational-assurance.md
Szczepanov: Added in 1119f3f: 6.2a now includes a `multi_event_taper_conflict_mid_horizon` case (authority/contributor transition on day 2–6) plus an explicit fix task — re-resolve objective admissibility for each day whose periodization phase differs from `todayDate`'s, and record the already-accrued-credit decision — instead of only a static scenario.

docs/plans/phase-6-evidence-and-operational-assurance.md
Szczepanov: Added in 1119f3f: 6.2b now includes a fix task — wire `FixedActivity.environment`/`equipment` into `resolveAvailability` and `reservedCapacityCost` into ranking/fatigue, or record an explicit decision for why they should stay unconsumed — instead of describing them only as contracts to observe.

docs/plans/phase-6-evidence-and-operational-assurance.md
Szczepanov: Fixed in 1119f3f: the `Addresses` line now explicitly names the two Phase-5 carryover gaps (5.6 mid-horizon re-resolution, 5.3 `FixedActivity` wiring) as owned by 6.2a/6.2b, separate from F11/F12/F15, and the task-board summary/primary-files columns and acceptance criteria were updated to match.

docs/plans/README.md
Szczepanov: Fixed in 1119f3f: both sentences now read "Phases 0–5 are implemented" / "Phases 0–5 are complete; Phase 6 remains Draft" instead of "all six phases," and the Phase 6 row's Addresses column now also names the Phase-5 carryover gaps.

Szczepanov:
Pushed 1119f3f addressing all review comments:

- **Wording accuracy (CodeRabbit + @Szczepanov):** F15 excluded from the fully-closed finding range; `docs/plans/README.md` and the completion review now say "Phases 0–5 are implemented," not "all six phases."
- **Verdict overstatement (@Szczepanov):** the completion review no longer implies full Phase 5 closure. It now names two Phase-5 residual gaps explicitly — the mid-horizon multi-event objective re-resolution gap in `planner.ts`, and the unconsumed `FixedActivity.environment`/`equipment`/`reservedCapacityCost` in `schedule.ts` — as *not* folded into F11/F12/F15, and hands them to Phase 6 rather than reopening Phase 5's `Implemented` status.
- **Phase 6 plan now owns real fixes, not just observation (@Szczepanov):** 6.2a gains a mid-horizon transition scenario plus a task to re-resolve objective admissibility across it; 6.2b gains a task to wire `FixedActivity` fields and `reservedCapacityCost` into availability/ranking; the `Addresses` line, task board, and acceptance criteria were all updated to reflect this ownership.
- **Implemented plans reading like work lists (CodeRabbit):** retitled the Pre-implementation-state/Work items/Tests/Fix/Current-state sections in `phase-1-live-defects.md` and `phase-2-plan-intent-authority.md` as historical/as-delivered, and rewrote the imperative "Generalise `event-plan.ts`" line as a completed outcome.

Replied inline on each thread with the specific fix. Resolving the threads now.

Szczepanov (commented):
Second-pass review completed after re-reading the Phase 5 integration code, the original F11/F12/F15 findings, and the current simulation/CI path.

### What I changed in this pass

I turned Phase 6 from a broad Draft into an executable, dependency-ordered plan and implemented the first safe slice (6.1):

- `simulate:scenarios` no longer overwrites `docs/analysis/simulation-baseline.json`;
- added `simulate:update-baseline -- --reviewed`, which regenerates the scenarios/gate before writing a normalized baseline and refuses to run without the explicit review flag;
- `simulate:diff` now fails clearly for a missing/malformed baseline and once again compares independently against the committed baseline;
- CI now asserts that a normal scenario run leaves the baseline byte-unchanged;
- `docs/plans/phase-6-evidence-and-operational-assurance.md`, the completion review, and the plan index now describe Phase 6 as **In progress** with explicit task dependencies, tests, acceptance criteria, rollback conditions, and recommended PR slicing.

### Corrected interpretation of the fixed-activity carryover

On the first review I described `FixedActivity.environment` / `.equipment` as fields that should be wired directly into `resolveAvailability`. After re-reading the model contract, I think that is too coarse: those fields describe the **fixed activity's venue**, not necessarily the athlete's whole-day availability. An outdoor football match should not make a separate morning indoor-bike session impossible.

The plan now separates three concerns:

1. `durationMin` / `availabilityOverride` reserve time;
2. explicit `expectedCost` / `expectedStimulus` make the fixed activity a projected exposure, affecting same-day capacity, objective credit, and next-day fatigue exactly once;
3. a true travel/day-wide equipment or environment restriction gets its own explicit day-context override rather than overloading activity venue metadata.

Missing expected cost is intentionally zero/unknown rather than the current undocumented `0.2` fallback becoming a live recommendation heuristic. `fixed:false` also remains on its authored date while greedy planning is production; automatic relocation belongs to any later sequence-search adoption.

### Outstanding work, in order

**6.2 — correctness first (behavior-changing, policy-version bump required)**

- 6.2a: re-resolve multi-event objectives on the date taper/contribution windows actually change inside the 7-day forecast. Reconciliation is by `ObjectiveKey`; already-earned/projected credit is preserved, but future admissibility and qualification change from that date onward. Add transition-on-day-2–6 tests and fresh-plan-vs-projected-day parity.
- 6.2b: project fixed activities as explicit training exposures. Apply authored cost/stimulus once, do not double-count completed activities, do not invent load when expected cost is absent, and add explicit day-context semantics for true travel constraints.

**6.3 — scenario contract**

Extend scenarios with `events`, `initialHistory`, `fixedActivities`, and date-level readiness. Add static + mid-horizon multi-event cases, football/fixed-load case, travel override, external-load/green-readiness, readiness crash/return, and inferred/partial completion. Contract-vs-observation classification belongs to individual checks, not the entire scenario.

**6.4 — F11 evidence layer**

Add daily decision traces plus a deterministic synthetic boundary corpus and trigger-frequency report. This is policy-regression/calibration evidence, not physiological validation and not an automatic tuning system.

**6.6 — coverage visibility**

Publish frontend/backend coverage summaries/artifacts. No arbitrary repository-wide percentage threshold until a concrete escaped defect justifies a targeted gate.

**6.5 — Firestore deployment assurance**

Blocked only by identifying the production Firebase project/deployment owner. Then choose either repository-owned protected deployment or read-only drift detection against the external owner. Emulator correctness remains separate from deployed-rule identity.

**6.7 — fatigue fusion last**

Keep production `max(external, internal)` until 6.3/6.4 expose a concrete undesirable trajectory. Compare candidates simulation-only; adoption requires contract-safe evidence, ADR, policy-version bump, reviewed baseline update, and rollback condition. Retaining `max()` is a valid outcome.

### Validation

Current head `ae16f0c4`: GitHub Actions CI Pipeline #180 is green. 486 frontend tests pass; 26 Firestore emulator tests pass; policy-drift check passes; 11 scenarios pass and leave the baseline unchanged; semantic diff reports **no differences**; production build, Python lint/type/tests and Docker build pass.

One non-blocking maintenance item remains visible rather than hidden: `npm audit` reports five moderate transitive advisories in the Firebase tooling chain; CI currently blocks only high-severity findings.

I do not see another blocker in the Phase 6.1 implementation itself. The recommendation-behavior changes in 6.2 should be separate PRs as specified in the plan, so each gets its own policy-version bump and semantic diff.

Szczepanov:
Pushed 0f73887: closed both Phase 5 correctness carryovers (task 6.2 from the phase-6 plan).

**6.2a — multi-event objectives now re-resolve per day.** `reconcileObjectivesForDate` (new, `planner.ts`) rebuilds each projected day's objective definition the same way a fresh plan on that date would, then reconciles it onto the running credit ledger by `ObjectiveKey` (D6-A: survived keys keep their credit, dropped keys' credit is remembered and restored if the key re-enters later in the horizon, new keys start at zero). This also fixes same-event Build→Specificity/taper transitions, not just multi-event ones — several scenarios' weekly `threshold_quality` target correctly drops once a mid-week taper stops demanding it.

**6.2b — fixed activities are now projected exposures, not just calendar blockers.** `schedule.ts`'s `reservedCapacityCostProfile` (dimensional, D6-C: no invented default) is consumed by same-day ranking and, at end of day, folded into next-day fatigue and objective credit via the same canonical credit primitive as a structured exposure. A new explicit `FixedActivity.availabilityContextOverride` gives true day-wide restrictions real semantics without letting an activity's own venue metadata (D6-B) leak into an unrelated session.

**Verification:** `npm run check` (502 tests), `npm run test:rules` (26 emulator tests), `simulate:scenarios`/`simulate:diff` (baseline reviewed and updated — every semantic change explained in the completion review), `compare:sequence-search` (0 violations), `uv run pytest/ruff/mypy`, and `check-policy-drift.mjs` all pass. `POLICY_VERSION` bumped to `2026-08-phase6-correctness-carryovers-v1`.

Docs updated to match: phase-6 plan's 6.2/6.2a/6.2b marked done with "Delivered as" notes, `docs/plans/README.md`'s Phase 6 row, and the completion review's finding table/remaining-work list.

app/src/engine/planner.ts
```diff
@@ -605,11 +707,80 @@ export function generateWeekAheadPlan(
...
+                    modality: 'None',
+                    earnedCredit: allocated,
+                });
+            });
+            microcycle = projected.microcycle;
+        });
+    };
+
     applyPick(todayDate, todayRec.template);
+    applyFixedActivityExposures(todayDate);
```
Szczepanov: **[Major] Fixed activities still cannot constrain the live or tomorrow recommendation.** `todayRec` and `tomorrowRec` are selected before `generateWeekAheadPlan` receives `fixedActivities` (the production `Home` path calls `evaluateTrainingWithIntent` / `evaluateNextDayPlanWithIntent`, neither of which accepts the fixed-activity set). Here we only apply the fixed exposure *after* those already-chosen picks. That means a booked football match or a true travel-day override on today/tomorrow can coexist with an otherwise impossible hard session, while only day 2+ gets the new availability/fatigue gates. Please thread fixed-activity availability into the actual today/next-day selection path (or recompute those recs through the same contract), and add production-path regressions for fixed activities on today and tomorrow.

app/src/engine/models.ts
```diff
@@ -132,6 +132,18 @@ export interface FixedActivity {
...
     availabilityOverride?: number;
+    /** Phase 6.2b / D6-B: a TRUE day-wide restriction, deliberately separate from this
+     *  activity's own `environment`/`equipment` above. An outdoor football match's venue
+     *  does not imply a separate same-day indoor session is also outdoor-only -- only set
+     *  this when the day itself is actually constrained (e.g. travel: every session that
+     *  day really is stuck at a hotel gym). Absent = this activity's venue never restricts
+     *  another session on the same date. Not yet covered by `firestore.rules`' `hasOnly`
+     *  allow-list (see 5.3's storage table) -- persistence validation is a follow-up
+     *  required before the UI can write this field. */
+    availabilityContextOverride?: {
```
Szczepanov, line 145: **[Major] `availabilityContextOverride` is not reachable through the production persistence path yet.** `validateFixedActivity` neither validates nor copies this field, so `FixedActivityService` drops it on both reads and writes; `firestore.rules` also excludes it from the fixed-activity `hasOnly(...)` allow-list, so a client write containing it is rejected. The new planner test bypasses those boundaries with a hand-built typed object (and even uses `durationMin: 0`, which persisted fixed activities reject). Please wire this field through validation + rules + service/UI persistence and add service/emulator coverage before marking the travel-day part of 6.2b complete.

app/src/engine/planner.ts
```diff
@@ -753,6 +944,7 @@ export function generateWeekAheadPlan(
         const pickCredits = creditingObjectivesFor(pick.template);
         const addressed = pickCredits.map(item => item.objective.title);
         applyPick(date, pick.template, pickCredits);
+        applyFixedActivityExposures(date);
```
Szczepanov: **[Major] The booked activity's projected stimulus is credited too late to prevent redundant same-day work.** `unresolved` is captured and the optional session is ranked/selected before this call; only afterward does `applyFixedActivityExposures` add `expectedStimulus` credit. So an evening fixed session that already satisfies `strength_maintenance`/another eligible objective can still cause the planner to prescribe an additional session for that same unresolved objective, then mark it resolved after both are scheduled. Apply the fixed activity's projected *stimulus credit* before ranking (while keeping its cost reserved/not-yet-completed), and add a regression that a booked objective-satisfying activity suppresses redundant same-day work.

app/src/engine/planner.ts
```diff
@@ -622,39 +793,54 @@ export function generateWeekAheadPlan(
...
         );
+        // Phase 6.2b: fold today's reserved (not-yet-happened) fixed-activity load into the
+        // fatigue signal used for THIS DAY's ranking/gating only -- externalFatigue itself
+        // (the carried-forward ledger) is untouched here, so this never marks the load as
+        // already completed before the activity occurs (D6-B/6.2b item 2). Its cost only
+        // becomes real, carried-forward load at the end of the day, via
+        // applyFixedActivityExposures below.
+        const rankingFatigueForDate: FatigueState = {
+            ...rankingFatigue,
+            combinedFatigue: combineMax(rankingFatigue.combinedFatigue, availability.reservedCapacityCostProfile),
```
Szczepanov: **[Major] Reserved fixed-activity cost is being fused with `max()` instead of added to projected external load.** `combineMax` is appropriate for external-vs-internal response fusion, but this reservation is *additional external load*. With existing lower-body fatigue 0.6 and a booked activity cost 0.5, this produces `max(0.6, 0.5) = 0.6`, reserving no additional lower-body capacity; the completed-load path would add then clamp toward 1.0. Please project the reserved cost onto a copy of the external-load state using the same additive/clamped semantics as `applyCompletedSessionLoad`, then fuse projected external with internal response. Add a test with non-zero pre-existing fatigue so this masking case is exercised.

app/src/engine/planner.ts
```diff
@@ -530,6 +530,99 @@ export function prepareWeekAheadPlanSeed(
...
+    microcycle.objectives.forEach(objective => {
+        if (!freshKeys.has(objective.key)) {
+            creditMemory.set(objective.key, snapshotObjectiveCredit(objective));
+        }
+    });
+
+    const objectives = fresh.objectives.map(definition => {
+        const existing = existingByKey.get(definition.key);
+        const carried = existing ? snapshotObjectiveCredit(existing) : creditMemory.get(definition.key);
+        if (!carried) return definition; // a key this projection has never seen: start at zero
```
Szczepanov: **[Major] A newly admitted objective starts at zero even when qualifying exposure already exists earlier in the rolling window.** This breaks 6.2a's stated “fresh plan on day N from equivalent projected history” contract: a fresh day-N build creates the newly active objective and replays relevant prior history through it, but this path returns a zero-credit definition whenever the key never existed earlier in this projection. Example: a contributor enters its 35-day window on day 4 after a qualifying day-1/2 session; the week-ahead path can prescribe redundant work that a fresh day-4 plan would consider already credited. Please seed genuinely new keys from the rolling completed/projected exposure ledger (or explicitly narrow the equivalence contract) and add a contributor-enters-window regression with prior qualifying exposure.

Szczepanov (commented):
Re-reviewed only the new delta since my previous pass (`ae16f0c4` → `0f738870`). CI #182 is green (502 unit tests, 26 Firestore-rule tests, policy-drift check, scenarios, build, Python and Docker), and the 6.2 implementation closes a meaningful part of the previously documented carryovers. However, I found five remaining correctness/integration gaps below that current tests do not cover, so I would not yet treat 6.2b / the day-N equivalence claim as fully closed.

I deliberately did not duplicate CodeRabbit's existing architecture-doc comment on `schedule.ts`.

One non-code cleanup item: the PR description's Validation section is stale — it still says head `ae16f0c4`, CI #180 and 486 tests; current head is `0f738870`, CI #182 and 502 tests.

Szczepanov (commented):
Recommendation-quality review against the current engine plus the supplied cycling macrocycle uncovered a deeper modeling problem that Phase 6 does not currently own. The athlete report that triggered this review is concrete: hard race-specific cycling on day -1, currently in Specificity, then the next-week recommendation collapses to essentially one race-specific session + Cycling Pedalling Economy + Yoga/Mobility. Immediate recovery/support after the hard ride can be correct; that whole-week structure is not consistent with the documented weekly architecture.

1. **Separate adaptation credit from weekly role/coverage fulfillment.** `objectivesFromDemand()` gives `zone2_aerobic` no category/intensity qualification. The enriched profiles give Hard Endurance and Race-Specific Endurance substantial `aerobicEndurance`, so those sessions can discharge the Zone-2 ledger. The committed simulation baseline demonstrates this directly: `end_hard_02` and `end_race_specific_01` receive `zone2_aerobic` credit. Physiologically that cross-adaptation is reasonable, but as a planning contract it is wrong: a hard race simulation should not replace the week's deliberately low-intensity aerobic volume. Keep fractional stimulus/adaptation credit, but add an orthogonal role/coverage ledger (or qualification) so `easy_aerobic` can only be fulfilled by the approved easy/recovery cycling role unless an explicit substitution policy says otherwise.

2. **The same collapse exists for sustained quality vs race-specific work.** `threshold_quality` currently requires threshold stimulus + allowed modality, not the `sustained_quality` role/category, so a race-specific session can partially/fully discharge the week's threshold requirement. Yet `event-plan.ts` already declares `sustained_quality`, `outdoor_event_specific`, `easy_aerobic`, `primary_strength`, etc. as distinct coverage keys. `generateWeeklyObjectives()` ignores that semantic distinction when creating the live ledger: in the plan-derived branch it switches on `ObjectiveKey`, while `coverageKey` is effectively metadata. Make required coverage keys/session roles first-class weekly constraints. One workout may earn adaptation credit on several axes, but it should not silently satisfy several distinct required session roles.

3. **The authored cycling plan is not actually generic to the athlete's target date.** `resolvePlanDefinitionForEvent()` only returns `buildSeptemberCyclingEventPlan()` for a cycling event whose date is literally `2026-09-20`, and that plan contains absolute Aug/Sep block dates. Any real event on another September date falls back to the generic demand-derived objective path — exactly the path with the role-collapse above. Generate the block schedule relative to `event.date` (or persist a real user-authored `PlanDefinition`) rather than keying the richer plan to one literal event date.

4. **Add the escaped case as a Phase-6 contract, not merely an observation.** Add a deterministic `cycling_specificity_after_hard_race_specific` scenario: A-priority cycling event in Specificity, exact hard Race-Specific Cycling exposure on day -1, otherwise healthy/available athlete. The next seven days should demonstrate: (a) no incompatible quality within the recovery/spacing window; (b) at least one *true Easy Endurance/Zone-2* cycling exposure; (c) at least one distinct sustained threshold/over-under quality role once recovered; (d) event-specific work appropriately spaced/repeated according to the rolling weekly contract; and (e) support/recovery sessions may appear but cannot replace the required developmental roles. This should be a coaching-policy contract. Phase 6.3 currently says exact category distributions remain observational; that is fine for exact counts, but **fulfillment of coverage keys already declared `required` cannot remain observational**.

5. **Strengthen the golden-week contract.** `goldenWeek.test.ts` currently accepts any two dates from Hard/Moderate/Race-Specific Cycling as the 'two key sessions', plus threshold/strength objective resolution and one recovery day. It therefore cannot distinguish `threshold + race-specific + Zone2` from two sessions that cross-credit the same objectives. Add separate assertions for realized `sustained_quality`, `outdoor_event_specific`, and true `easy_aerobic` coverage, and add a Specificity variant seeded with prior-day hard work.

6. **Do not solve this first by tuning fatigue thresholds or simply enabling beam search.** ADR-0015's beam prototype is promising (better objective resolution, fewer recovery days with no added hard-gate violations in its existing harness), but it optimizes the same objective/coverage semantics. Fix the weekly contract first, then rerun greedy vs beam on the new escaped-case scenario. If beam then materially improves role fulfillment with acceptable single-call latency, that is much stronger evidence for revisiting the deferred adoption decision.

This is more than subjective dislike of a recommendation: the current implementation has conflated *cross-adaptation* with *session-role substitution*. The former is useful physiology-aware accounting; the latter needs an explicit coaching policy. I would add this as a new Phase 6 correctness item before calibration/fatigue-fusion work and not consider the recommendation engine behaviorally complete until the Specificity regression passes.

app/src/engine/schedule.ts
```diff
@@ -40,16 +55,65 @@ function resolveOwnedEquipment(constraints: UserContext['constraints'] | null |
...
+    }
+
+    let equipment: string[] | null = null;
+    for (const o of overrides) {
+        if (!o.equipment) continue;
+        equipment = equipment === null ? o.equipment : equipment.filter(item => o.equipment!.includes(item));
+    }
+
+    return { environment, equipment };
 }
```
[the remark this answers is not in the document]
Szczepanov: Fixed in 860f7a3: added the Phase 6.2b/D6-B rule (`availabilityContextOverride` vs. activity-only `environment`/`equipment`) plus a full description of the projected-exposure semantics (reserved cost, stimulus credit ordering) to `docs/architecture/recommendation-engine.md`'s Fixed activities section.

app/src/engine/planner.ts
Szczepanov: Agreed, this was a real gap. Fixed in 860f7a3: `evaluateTrainingWithIntent`/`evaluateNextDayPlanWithIntent` (`rules.ts`) now accept `fixedActivities` and apply the same availability/reserved-cost-fatigue/stimulus-credited-before-ranking treatment as the week-ahead loop, scoped to that call's own date. `buildOptimizationContext` (`optimizer.ts`) now takes `fixedActivities` directly rather than always resolving availability with an empty list internally. `Home.tsx` now fetches today's/tomorrow's fixed activities and passes them into both calls (fails open to `[]` with a warning on a read failure, unlike the week-ahead strip's fail-closed forecast -- this is an interactive one-day decision). Added a "Phase 6.2b" describe block in `trainingIntentAcceptance.test.ts" covering today's own availabilityOverride/stimulus-credit effect on `evaluateTrainingWithIntent` and tomorrow's own activity's effect via `evaluateNextDayPlanWithIntent`. Cross-day cost propagation from a still-uncompleted PRIOR-day activity remains out of scope (each call independently replays real completed history via `resolveTrainingIntent`) -- noted explicitly in the code and the phase-6 plan's amendment section.

app/src/engine/models.ts
Szczepanov: Fixed in 860f7a3: wired `availabilityContextOverride` through both persistence-boundary validators -- `validateFixedActivity` (`validation.ts`) now validates and copies it, and `firestore.rules`' `hasValidFixedActivity` now allows and validates it via a new `hasValidAvailabilityContextOverride`, mirrored between the two. `FixedActivityService` already round-trips whatever `validateFixedActivity` produces (every read/write already goes through it), so no separate service change was needed. Also fixed the planner test's `durationMin: 0` (a real persisted activity requires `durationMin > 0`) to `durationMin: 1`. Added coverage in `validation.test.ts` (accept/reject cases) and the Firestore emulator suite (accept, reject-unknown-environment, reject-unrecognized-key).

app/src/engine/planner.ts
Szczepanov: Confirmed and fixed in 860f7a3. `applyFixedActivityStimulusCredit`/`fixedActivityCostProfileForDate` are now module-level pure functions (`planner.ts`) called in the correct order: stimulus credit BEFORE `unresolvedObjectives` is computed for that day's ranking (both in the week-ahead loop and in `rules.ts`'s `evaluateTrainingWithIntent` for today's own activities), cost AFTER the pick (it only becomes real once the day has happened). Added a regression test in `planner.test.ts` -- a booked activity that fully resolves `strength_maintenance` measurably lowers a same-day Strength candidate's utility via optimizer.ts's `isStrengthResolved` suppression, which only fires correctly with this ordering -- plus a live-path version in `trainingIntentAcceptance.test.ts` checking the actual `candidateScores` in `decisionTrace`.

app/src/engine/planner.ts
Szczepanov: Confirmed and fixed in 860f7a3. Replaced `combineMax(rankingFatigue.combinedFatigue, reservedCapacityCostProfile)` with `applyCompletedSessionLoad(rankingFatigue, date, reservedCapacityCostProfile)` (elapsedHours=0 since `rankingFatigue` is already decayed to `date`, so it adds without further decay) -- reuses the exact additive-then-clamped-then-max-with-internal-response semantics a real completed session already gets, transiently for ranking only (never written back to `externalFatigue`). Applied identically in `rules.ts`'s `evaluateTrainingWithIntent`. Added the test you asked for: non-zero pre-existing lower-body fatigue (~0.54, already exceeding the 0.5 reserved cost -- exactly the case `max()` gets wrong) plus a same-day booked activity, asserting `peakFatigue` increases by more than `max()` could ever produce.

app/src/engine/planner.ts
Szczepanov: Confirmed and fixed in 860f7a3. `planner.ts` now tracks a `projectionExposures` ledger (every applied pick's/fixed-activity's own stimulus, dated) and `reconcileObjectivesForDate` replays exposures strictly before the reconciled date against a newly-admitted objective definition via a new `backfillCreditFromPriorExposures`, capped at the objective's required amount and recorded as `projectedCredit` (not `completedCredit`, since it's still a projection). Added your exact example as a regression test in `planner.test.ts`: a contributor entering its 35-day window on day 4 after a qualifying day-1 exposure now backfills credit instead of starting at zero, plus a sibling test confirming an exposure dated on/after the reconciled date is correctly NOT backfilled.

Szczepanov:
Pushed 860f7a3 addressing all 6 review comments — the 5 substantive findings were real bugs in the previous Phase 6.2 commit, not style notes, and are fixed properly rather than patched around:

1. **Today/tomorrow never saw fixed activities at all** — `evaluateTrainingWithIntent`/`evaluateNextDayPlanWithIntent` now accept `fixedActivities` and apply the same treatment the week-ahead loop does; `Home.tsx` fetches and passes them; `buildOptimizationContext` now resolves availability from a single source of truth instead of silently discarding fixed activities internally.
2. **`availabilityContextOverride` wasn't persistable** — wired through `validateFixedActivity` and `firestore.rules`, mirrored between both; fixed the test that relied on an activity shape a real write would reject.
3. **Stimulus credit landed after ranking, not before** — reordered so a booked activity that already satisfies an objective actually prevents redundant same-day work, with a regression test proving the optimizer's own suppression mechanism now fires.
4. **Reserved cost was fused with `max()`, masking it under real fatigue** — switched to the same additive/clamped semantics a completed session gets, with a test using non-zero pre-existing fatigue (the exact case that broke before).
5. **A newly-admitted objective always started at zero credit** — now backfills from qualifying same-projection exposures dated before the reconciled day, matching the "fresh plan on day N" contract; used your own example as the regression test.

Verification: `npm run check` (515 tests), `npm run test:rules` (29 emulator tests), `simulate:scenarios`/`simulate:diff` (no drift — none of the 11 committed scenarios exercise `fixedActivities`), `compare:sequence-search`, Python suite, `check-policy-drift.mjs`. No new `POLICY_VERSION` bump needed since these are corrections within the same not-yet-merged 6.2 change.

Replied inline on each thread with specifics. Resolving now.

app/src/engine/planner.ts
```diff
@@ -530,6 +530,244 @@ export function prepareWeekAheadPlanSeed(
...
+ * run's own first-party pick or authored fixed-activity stimulus, not external evidence. */
+function backfillCreditFromPriorExposures(
+    definition: WeeklyObjective,
+    priorExposures: readonly ProjectionExposure[],
+): number {
+    const requiredCredit = definition.requiredCredit ?? definition.targetExposures;
+    let total = 0;
+    for (const exposure of priorExposures) {
+        if (total >= requiredCredit) break;
+        const credit = deriveObjectiveCreditFromProfile(definition, exposure.stimulus, {}, {
```
Szczepanov, line 504: **[Major] Prior-exposure backfill now makes the adaptation-credit/session-role conflation persistent across the horizon.** This replay is mechanically correct, but `definition` is still the generic `WeeklyObjective`: `zone2_aerobic` has no category/intensity qualification, and `threshold_quality` only requires threshold stimulus + Cycling modality. Therefore a prior hard Race-Specific ride can be replayed here and satisfy credit that the authored plan intends as separate `easy_aerobic` and `sustained_quality` coverage. That is exactly the escaped recommendation pattern we discussed. Please separate adaptation credit from required weekly role/coverage fulfillment (or add role/category/intensity qualification strong enough to preserve those roles) before treating this backfill as fulfillment authority. Add the regression: hard race-specific Cycling on day -1 in Specificity must not eliminate the upcoming true Zone-2 exposure or distinct sustained threshold/over-under role once recovered.

app/src/engine/planner.ts
```diff
@@ -530,6 +530,244 @@ export function prepareWeekAheadPlanSeed(
...
+    let nextMicrocycle = microcycle;
+    const credits: PlannedObjectiveCredit[] = [];
+    const exposures: ProjectionExposure[] = [];
+
+    dayActivities.forEach(activity => {
+        const stimulus: WorkoutStimulusProfile = { ...ZERO_STIMULUS, ...activity.expectedStimulus };
+        exposures.push({ date, stimulus, modality: undefined, category: undefined });
+
+        const derivedCredits = getUnresolvedObjectives(nextMicrocycle, true).flatMap(objective => {
+            const credit = deriveObjectiveCreditFromProfile(objective, stimulus);
```
Szczepanov: **[Major] A FixedActivity's expected stimulus still has no activity identity, so objective credit can be both under- and over-attributed.** `deriveObjectiveCreditFromProfile(objective, stimulus)` receives no modality/category context because `FixedActivity` has neither field. A booked Cycling workout therefore cannot credit modality-scoped cycling objectives, while a football activity with enough aerobic stimulus can still credit the unqualified `zone2_aerobic` objective. Please persist an explicit activity modality/category (or a linked template/coverage role) when expected stimulus is intended to earn objective credit, and pass that identity into the canonical credit primitive. This should be tested with both a booked cycling-specific session that *does* credit the right cycling role and football that does *not* satisfy the week's true easy-aerobic cycling coverage.

app/src/engine/rules.ts
```diff
@@ -975,11 +1022,14 @@ export async function evaluateNextDayPlanWithIntent(
...
     preparedHistorySnapshot?: TrainingHistorySnapshot | null,
+    /** Phase 6.2b: threaded straight through to `evaluateTrainingWithIntent` for tomorrow's
+     *  date -- see that function's own doc comment. */
+    fixedActivities: FixedActivity[] = [],
 ): Promise<NextDayPotentialPlan> {
     const scenarios = buildNextDayScenarios(todayReadiness, context, todayDate, todayRec);
     const evaluate = async (scenario: NextDayScenario) => evaluatedBranch(
         scenario,
-        await evaluateTrainingWithIntent(userId, scenario.readiness, context, events, scenarios.date, todayRec.mode, historyProvider, preparedHistorySnapshot),
+        await evaluateTrainingWithIntent(userId, scenario.readiness, context, events, scenarios.date, todayRec.mode, historyProvider, preparedHistorySnapshot, fixedActivities),
```
Szczepanov: **[Major] Tomorrow's provisional intent still ignores today's selected session as projected history/load/credit.** This call resolves tomorrow from the prepared completed-history snapshot plus hypothetical readiness, but `todayRec` itself is only used by `buildNextDayScenarios` to adjust the hard-session count. Its stimulus is not credited into tomorrow's microcycle, its cost is not applied to tomorrow fatigue, and it is absent from recovery-spacing history. `generateWeekAheadPlan` later consumes this already-selected `tomorrowRec`, so it cannot repair a redundant/too-aggressive day-1 pick after applying today's session. Please evaluate tomorrow from a projection that includes today's chosen recommendation (cost + stimulus + role/history), or recompute the provisional pick inside the same stateful projection. Add a regression where today's race-specific/hard session changes tomorrow's objective state and spacing.

Szczepanov (commented):
Re-review of current head `860f7a3c` after the latest fixes.

The good news: the five correctness gaps from my `0f738870` review are materially addressed here (today/tomorrow fixed activities, persisted day-context override, pre-ranking fixed stimulus credit, additive reserved cost, and new-objective backfill). CI #183 is green.

However, the later recommendation-quality review (submitted at 10:42 UTC) is still largely unimplemented. The current code still conflates adaptation credit with required weekly session-role coverage, and the richer cycling plan still only resolves for the literal event date `2026-09-20`. That is especially important because the supplied macrocycle targets the race around Sep 5–6 or Sep 12–13, so the athlete's actual event falls through to generic demand-derived objectives rather than the authored `easy_aerobic` / `sustained_quality` / `outdoor_event_specific` coverage contract.

I also found two concrete integration defects below: the provisional tomorrow path does not project today's selected recommendation into tomorrow's fatigue/objective/history state, and fixed-activity stimulus has no modality/category identity so it cannot correctly distinguish a booked cycling session from football while awarding objective credit.

Separately, CodeRabbit's current Major comment on `rules.ts` is valid: today's check-in time budget is used to build `candidates`, then discarded because `buildOptimizationContext` resolves availability again with `checkin=null` and `rankCandidates` receives that second object. I did not duplicate that thread.

One more weekly-architecture concern remains even after the role ledger is fixed: `resolveWeeklyAnchors` nominates one quality date in a fatigue-independent pre-pass. If projected fatigue makes that exact date `modify`/`recover`, the quality role is simply missed; there is no later re-nomination. That is a plausible mechanism for the observed `race-specific + pedalling economy + mobility` week after a hard prior-day ride and should be covered by the escaped-case contract.

I would not merge this as recommendation-behavior complete yet. Fix the inline items plus the existing CodeRabbit Major, then add the Specificity escaped-case and make required coverage roles authoritative before tuning fatigue or revisiting beam-search adoption.

docs/plans/phase-6-evidence-and-operational-assurance.md
```diff
@@ -0,0 +1,674 @@
...
+One scenario can contain both a blocking safety assertion and useful observational
+metrics. Classification belongs to the assertion/check, while every scenario can still
+emit observations.
+
+Keep all dates as Europe/Warsaw calendar-date strings and use
+`addDaysToLocalDateString`; never derive fixture dates through UTC serialization.
+
+### Targeted cases
+
+Keep the existing 11 scenarios as controls and add at least these deterministic cases:
```
Szczepanov, line 385: **[Major] Phase 6.3 still does not own the escaped Specificity recommendation that triggered this review.** Please add a deterministic `cycling_specificity_after_hard_race_specific` case: A-priority cycling event in Specificity, exact hard Race-Specific Cycling exposure on day -1, otherwise healthy/available. Make these **contracts**, not distribution observations: recovery/spacing immediately after the hard ride; at least one true Easy Endurance/Zone-2 cycling exposure in the next 7 days; a distinct sustained threshold/over-under role once recovered; appropriately spaced event-specific work; support/technical/recovery sessions may appear but cannot substitute for required developmental roles. This scenario should also fail if a pre-nominated quality anchor is fatigue-gated and never re-nominated later in the horizon.

Szczepanov:
I added a detailed implementation plan for the recommendation-quality findings from the latest review:

[`docs/plans/phase-6-2c-recommendation-quality-and-weekly-coverage.md`](https://github.com/Szczepanov/adaptive-training-recommender/blob/codex/phase-0-5-completion-audit/docs/plans/phase-6-2c-recommendation-quality-and-weekly-coverage.md)

Commit: `4348685d48f6f7a722d61f06e9e2c33cd2612351`.

### Core conclusion

The primary defect is not a bad fatigue coefficient. The engine currently uses the same `WeeklyObjective` ledger for two different concepts:

1. physiological/adaptation credit; and
2. weekly programming-role fulfillment.

That lets a hard race-specific ride legitimately earn aerobic/threshold stimulus and then incorrectly behave as though it also fulfilled the week's true Zone-2 and sustained-quality roles. The Phase 6.2a projected-history backfill is correct, but it makes this underlying semantic problem propagate more consistently.

### Proposed architecture

The plan proposes **two orthogonal ledgers**:

- adaptation credit remains fractional and stimulus-based;
- weekly coverage is count/role-based and comes only from explicit event-plan/catalog mappings (`easy_aerobic`, `sustained_quality`, `outdoor_event_specific`, etc.).

A session may earn several adaptations, and may fulfill several coverage roles only when the authored coverage table explicitly maps it to those roles. Stimulus overlap alone never creates a role substitution.

I recommend recording this as ADR-0016 before implementation.

### Implementation order

1. Fix the current `ResolvedAvailability` split-brain so today's check-in time budget reaches the same hard gate used by ranking.
2. Add the coverage-state primitives and exact template/workout -> coverage mapping.
3. Replay completed/projected history into adaptation and coverage separately.
4. Make required coverage first-class in ranking using an ordinal/lexicographic `coverageNeedTier`, not another arbitrary scalar multiplier.
5. Replace the literal `2026-09-20` cycling plan with an event-relative plan aligned to the existing Build / Specificity / taper boundaries; travel becomes an availability overlay, not a fabricated default phase.
6. Add structured identity to fixed activities so known catalog sessions can earn correct modality/category/coverage credit while unknown activities fail closed for coverage.
7. Project today's selected recommendation **and today's still-uncompleted fixed load** into tomorrow's provisional green/yellow/red branches with correct fatigue decay, projected objective/coverage credit, and spacing history.
8. Make quality/event-specific anchors repairable: if projected fatigue makes a nominated anchor infeasible, re-nominate the unresolved role to a later safe slot rather than silently losing it.
9. Add `cycling_specificity_after_hard_race_specific` as a blocking synthetic coaching contract and strengthen `goldenWeek.test.ts` to assert true `easy_aerobic`, `sustained_quality`, and `outdoor_event_specific` coverage rather than broad key-session counts.
10. Only after those semantics are green, re-run greedy vs beam and judge beam on coverage fulfillment + safety + latency. ADR-0015 remains unchanged unless new evidence justifies a separate adoption decision.

The plan includes concrete target types, file/symbol ownership, regression cases, PR slicing, policy-version/replay requirements, acceptance criteria, risks, and rollback.

I would treat this 6.2c correctness work as higher priority than Phase 6 calibration/fatigue tuning. The escaped recommendation should be impossible to reproduce once the blocking Specificity scenario is in CI.

Szczepanov:
Current review status after re-checking head `4c6bbfc4a5325692c175865e53e81a6d083f1136` and CI Pipeline #223.

## Bottom line

The main recommendation-quality architecture is now materially implemented, and CI is green. However, I would **not yet say every review comment is addressed**. There are four unresolved UI threads whose substantive requests are mostly implemented, but CodeRabbit's latest review also contains actionable outside-diff/nitpick findings that do not all appear in the normal unresolved-thread count.

### What is now implemented

- Physiological adaptation credit is separated from explicit weekly programming-role coverage. Coverage is count-based and requires exact authored workout/template identity rather than broad stimulus overlap.
- Cycling plans are event-relative; runtime planning no longer depends on a literal September race date. Test fixture dates can remain fixed for reproducibility, but they are not scheduling authority.
- `easy_aerobic`, `sustained_quality`, `outdoor_event_specific`, strength, and recovery are distinct coverage functions.
- Today's check-in-derived time budget is reused through ranking.
- Tomorrow planning now projects today's selected recommendation into history/load/stimulus with `recommendation:<date>` occurrence identity.
- Fixed activities can persist exact `templateId`/`workoutId` catalog identity and stimulus credit uses the resolved modality/category.
- Stable occurrence identities exist for recommendation/completed/fixed projections.
- The Specificity escaped-case (`cycling_specificity_after_hard_race_specific`) exists, with direct contract coverage for true easy aerobic, distinct sustained quality, event-specific work, spacing, and recovery/support.
- Beam branch pruning was changed to preserve lexicographic coverage-tier ordering.
- Current GitHub Actions CI Pipeline #223 is fully green: frontend tests, Firestore emulator, simulation scenarios/aggregate gate, semantic diff, frontend build, Python suite, and Docker build all pass.

## Remaining substantive issues

### 1. Major — exact fixed-activity coverage is still not fully propagated through live/tomorrow ranking

`evaluateTrainingWithIntent()` applies same-day fixed-activity **adaptation** stimulus before ranking, but `buildOptimizationContext()` still derives `coverageState` from `intent.history` / `recentHistory`. The unfinished fixed activity's exact `fixed:<activity-id>` occurrence/catalog identity is therefore not necessarily present in the coverage input for the live decision.

The tomorrow helper has the same integration gap: `fixedActivityProjection()` currently projects cost/stimulus but does not preserve `occurrenceKey`, `templateId`, `workoutId`, resolved modality, and category. That means an exact booked cycling workout can affect fatigue/adaptation while still failing to satisfy its explicitly authored weekly coverage role.

Required fix:

- add identity-resolved unfinished fixed activities to the coverage input used by the live optimization context;
- project tomorrow fixed activities with `fixed:<activity-id>` plus exact catalog identity/modality/category;
- regression: an exact fixed Zone-2 / event-specific cycling activity reduces the corresponding coverage urgency on both today and tomorrow;
- anonymous/football fixed activity must continue to fail closed for cycling coverage.

### 2. Data integrity — projected -> completed duplicate source is order-dependent

`buildCoverageState()` currently deduplicates by first-seen `occurrenceKey`. If the same occurrence is supplied first as `projected` and later as `completed`, the count is correctly one, but the retained credit source depends on input order.

This should be deterministic: aggregate by occurrence first and prefer `completed` whenever any representation of that occurrence is completed. Then assert:

- one credit only;
- `source === 'completed'`;
- `completedSessions === 1`;
- `projectedSessions === 0`.

The existing occurrence regression currently verifies count/idempotency but not this source upgrade.

## Remaining smaller review items

These are not equivalent to the two items above, but they are still actionable feedback present in CodeRabbit review bodies:

- Move `templateId` / `workoutId` declarations into the canonical `FixedActivity` interface in `models.ts` instead of relying on module augmentation in `fixedActivityIdentity.ts`.
- Extend fixed-activity identity tests with an unknown/inactive `workoutId` failure case (the workout-only happy path is now covered).
- Add direct availability test for conflicting same-day non-`either` environment overrides.
- Add the Firestore-emulator regression for `availabilityContextOverride.equipment` exceeding 20 items.
- Export `ObjectiveCreditSnapshot` if `reconcileObjectivesForDate` is intended as an exported/tested API.
- Record/tighten the measured stressed-vs-baseline tolerance rather than leaving the broad `baselineCredit * 1.2` unexplained.
- Remove/reconcile the duplicate availability resolution in the week-ahead per-day flow.
- `Home.tsx` still performs a separate today/tomorrow fixed-activity query overlapping the week-ahead range read; this is low severity but was explicitly reviewed.
- PR description validation section is stale (`ae16f0c4`, CI #180, 486 tests) and should be refreshed to the final head/run after the remaining changes.

## Thread bookkeeping

GitHub currently still shows four unresolved review threads from the recommendation-quality review. Their original substantive requests — dual ledgers, fixed-activity identity, today->tomorrow projection, and the Specificity contract — are now mostly implemented, so after the two remaining integration/integrity issues above are fixed and validated, those threads should be replied to with the validating commit/CI evidence and resolved rather than left stale.

## Merge recommendation

I would keep the PR open until:

1. exact fixed-activity coverage is propagated through live + tomorrow coverage state;
2. occurrence source resolution prefers completed deterministically;
3. the targeted regressions above are green;
4. remaining actionable review cleanup is either implemented or explicitly dispositioned;
5. stale review threads are replied to/resolved and the PR Validation section is updated.

I would **not** tune fatigue coefficients or revisit beam-search adoption as part of these remaining fixes. The current work is still semantics/integration correctness, and CI #223 shows the broader branch is otherwise healthy.

app/src/engine/planner.ts
Szczepanov: Confirmed addressed, verified by tracing the code directly (not just reading the diff). ADR-0016 / Phase 6.2c (`docs/adr/0016-adaptation-credit-and-weekly-coverage.md`, `docs/plans/phase-6-2c-recommendation-quality-and-weekly-coverage.md`) separates this exactly as you describe: `coverage.ts` is an independent, count-based ledger keyed by exact `workoutId`/`templateId` identity (never stimulus magnitude, title, or broad category), and `coverageNeedTierForTemplate`'s tier is the PRIMARY sort key in `rankCandidates` (`optimizer.ts`, checked before benefit tier and utility). A prior hard Race-Specific ride can still credit `zone2_aerobic`/`threshold_quality` in the generic `WeeklyObjective` ledger via `backfillCreditFromPriorExposures`, but that no longer has fulfillment authority over the weekly role -- `easy_aerobic`/`sustained_quality` stay unfulfilled in the coverage ledger until an exposure with the matching exact workout identity lands, which is what actually drives ranking now. Your exact regression is the named contract `cycling_specificity_after_hard_race_specific` (`specificityCoverageContract.test.ts`/`specificityScenario.test.ts`, 8/8 passing across the coverage test files). Scope note: this protection currently applies to cycling events with a resolved `PlanDefinition` (now event-relative via `buildCyclingEventPlan`, not September-specific) -- non-cycling event categories and the pure no-event path still use the generic ledger only, so the conflation you flagged can still occur there; noted in the phase-6-2c plan.

One important related finding from verifying this end-to-end: reviewing `npm run simulate:diff` after this work shows a rest/recovery-share spike (14%→50%) in several healthy cycling scenarios that the phase-6-2c plan's own rollback conditions (§10) flag as disqualifying. I traced it and it is NOT a bug in this coverage logic -- it's a pre-existing fatigue-model interaction (Mobility/Recovery's non-zero cost vs Rest's zero cost) that 6.2c's more consistent coverage-driven scheduling now triggers far more often. Full writeup in `phase-6-2c-recommendation-quality-and-weekly-coverage.md` §9. I have not updated the baseline pending that being resolved.

app/src/engine/planner.ts
Szczepanov: Confirmed addressed. `FixedActivity` can now carry optional `templateId`/`workoutId` (`fixedActivityIdentity.ts`, `models.ts`, service/rules boundary), and `applyFixedActivityStimulusCredit` (`planner.ts`) only lets `expectedStimulus` earn objective credit when that identity resolves to one exact catalog prescription -- a legacy/untyped activity (e.g. football with no linked workout) can still reserve time/cost but cannot invent credit for an unqualified axis, and a booked cycling-specific session with real identity correctly credits the matching cycling role. The same exact identity feeds `coverage.ts`'s `coverageKeysForExposure`, so it's also used consistently for weekly-role coverage, not just objective credit. Test coverage: `fixedActivityIdentity.test.ts`, `coverage.test.ts`'s exact-mapping case, and `fixedActivityService.test.ts`'s round-trip/mismatch-rejection cases.

app/src/engine/rules.ts
Szczepanov: Confirmed addressed. `projectedProviderForTomorrow` (`rules.ts`) now builds a synthetic `TrainingHistoryProvider` that merges real trailing history with today's chosen recommendation (via `recommendationProjection`) and today's uncompleted fixed activities before `evaluateNextDayPlanWithIntent` calls `evaluateTrainingWithIntent` for tomorrow -- so tomorrow's `resolveTrainingIntent` call replays today's session's exact stimulus/cost/role through the normal history-replay path (`buildMicrocycleState`/`buildFatigueStateFromHistory`) exactly as a real day-later evaluation would, rather than only adjusting the hard-session count via `buildNextDayScenarios`. `generateWeekAheadPlan` then still consumes the resulting `tomorrowRec` as given, same as before, but that recommendation itself is now computed with today's session already accounted for. Documented in the phase-6-2c plan §7 ("Today/tomorrow projection contract").

docs/plans/phase-6-evidence-and-operational-assurance.md
Szczepanov: Confirmed addressed. `cycling_specificity_after_hard_race_specific` exists in `simulation/scenarios.ts` (A-priority cycling event exactly 35 days out, exact hard Race-Specific Cycling exposure with real identity in completed history on day -1, otherwise healthy/available) and is asserted as a hard CONTRACT (not an observation) in `specificityCoverageContract.test.ts`: recovery/spacing immediately after the hard ride is allowed, but the rolling week must still show a true Easy Endurance/Zone-2 exposure, a distinct sustained-quality/Hard-Endurance role, and renewed event-specific work, with support/technical/recovery sessions unable to substitute for those developmental roles. It's also wired into the shared aggregate harness (`specificityScenario.test.ts` + `simulate:scenarios`/`simulate:diff`), so it's regression evidence going forward, not a one-off fixture. All 8 coverage/specificity contract tests pass.

Flagging the same blocking finding here since this is the file the checklist lives in: verifying this end-to-end surfaced a rest/recovery-share spike (14%→50%) in OTHER healthy cycling scenarios that trips this same plan's own rollback condition (§10). Root-caused (not a 6.2c logic bug -- a pre-existing Mobility/Recovery-vs-Rest cost/benefit interaction that 6.2c's coverage-driven scheduling now triggers more often) and documented in §9 with candidate fixes, but not resolved or baselined -- see the plan doc and my PR summary comment.

Szczepanov:
Pushed 6d11214. Read and verified all 4 remaining review comments against the current head (the 40 commits of Phase 6.2c / ADR-0016 work: independent adaptation-credit vs. weekly-coverage ledgers, exact fixed-activity/workout identity, event-relative cycling plans, today's pick projected into tomorrow's intent).

**All 4 comments are confirmed architecturally addressed** — replied inline on each with the specific code/test evidence:
1. Adaptation-credit/session-role conflation → `coverage.ts`'s independent, exact-identity-only ledger now has ranking authority (`coverageNeedTier` is the primary sort key), not the generic stimulus-credit backfill. Scoped to cycling events with a resolved `PlanDefinition` — noted as a residual gap for non-cycling events.
2. FixedActivity stimulus lacking identity → optional `templateId`/`workoutId` now gate whether `expectedStimulus` can earn credit at all.
3. Tomorrow ignoring today's session → `projectedProviderForTomorrow` now feeds today's chosen recommendation into tomorrow's `resolveTrainingIntent` call.
4. Missing escaped-Specificity contract → `cycling_specificity_after_hard_race_specific` exists with hard contract assertions, 8/8 passing.

Full verification: `npm run check` (537 tests), `npm run test:rules` (29 emulator tests), `uv run pytest/ruff/mypy` — all green.

**One important finding that blocks calling this fully done:** reviewing `npm run simulate:diff` (a required step per the phase-6-2c plan's own checklist) shows several healthy cycling scenarios' rest/recovery share jumping from ~14% to exactly 50%, with `cycling_criterium_stressed_A` losing `strength_maintenance` coverage entirely (3/4 → 0/4). This trips the plan's own rollback condition (§10) for "an unexplained recovery/rest spike," so I did **not** bless/update the baseline.

I traced it day-by-day rather than guessing: it is **not a bug in the 6.2c coverage/backfill/projection code** — all three behave exactly as specified when inspected directly. It's a pre-existing fatigue-model interaction that 6.2c's more consistent coverage-driven scheduling now triggers far more often: `Mobility/Recovery` templates carry a non-zero cost profile while `Rest` is zero-cost, and `calculateStimulusBenefit` scores `Mobility/Recovery` slightly higher than `Rest` regardless of the athlete's `preferActiveRecovery` setting — so once fatigue crosses the recover threshold, picking `Mobility/Recovery` instead of `Rest` partially offsets each day's decay, letting fatigue oscillate just above threshold for several consecutive days instead of clearing it. Full writeup with candidate fixes (gate the Mobility/Recovery preference behind `preferActiveRecovery`; prefer `Rest` once in recover tier; revisit the cost profile) is in `docs/plans/phase-6-2c-recommendation-quality-and-weekly-coverage.md` §9.

Per this repo's own D6-E principle, a fatigue-model calibration change needs deliberate evidence review, not an ad hoc patch from me — so I've documented the finding precisely and left the baseline untouched rather than either hiding it or guessing at a fix. I also linked 6.2c into the main phase-6 plan and `docs/plans/README.md` so it isn't an orphaned document, and updated its own status/checklist to reflect this accurately.

Resolving the 4 threads since their specific concerns are addressed — the recovery-spike finding is tracked separately in the plan docs and this comment, not hidden by resolving them.

Szczepanov:
## Post-analysis review — recommendation quality vs active macrocycle v5.0

I reviewed the current PR head (`6d11214`) against the active **Sustained Multidirectional Field Macrocycle v5.0 (3 Aug 2026)**, not only against the existing code/tests. CI #224 is green and the earlier review threads are resolved, but I would **not merge the recommendation-policy portion yet**. The PR itself already withholds semantic-baseline acceptance because of the rest/recovery spike; the macrocycle comparison surfaces two additional behavior-contract gaps.

### What is directionally strong

- Separating adaptation credit from weekly programming-role coverage is the right architecture.
- Exact workout/template identity for coverage/fixed activities is much safer than broad stimulus inference.
- Today→tomorrow projection, occurrence identity, fixed-activity cost/stimulus ordering, and explicit day-wide availability overrides close real planner gaps.
- Keeping the previous semantic baseline when a rollback condition fires is exactly the right discipline.

### 1. **Blocker: A-priority taper starts materially too early for this athlete/event**

`buildCyclingEventPlan()` currently uses:

```ts
const taperDays = event.priority === 'A' ? 14 : event.priority === 'B' ? 5 : 0;
```

That does not match the active macrocycle.

For the earlier likely race date (5–6 Sep), the macrocycle explicitly makes **24–30 Aug the decisive peak race-specific week** and starts taper on **31 Aug** (~5–6 days out). For the later race (12–13 Sep), **31 Aug–6 Sep remains an additional specific week** and taper begins **7 Sep**.

With the current A=14 rule, a 5 Sep planning date enters taper around **22 Aug**, so the planner can remove `sustained_quality`, `outdoor_event_specific`, surge/gap-closing development, etc. during the exact week the macrocycle says should contain the race-specific simulation and repeated-surge work.

**Recommendation:** do not derive taper duration from A/B/C priority alone. Make taper start/duration an explicit event/plan parameter (with a sensible sport-specific default), and encode this macrocycle's current contract as deterministic tests:

- earliest-date branch: 24–30 Aug is still peak/specificity; taper starts 31 Aug;
- later confirmed branch: 31 Aug–6 Sep remains specific; taper starts 7 Sep.

Priority should affect importance/competition protection, not silently impose a generic 14-day taper that overrides the authored macrocycle.

### 2. **Major: a Recovery Spin can satisfy the “true Zone 2” coverage floor**

`event-plan.ts` maps both:

- `cycling_zone2_standard_01`, and
- `cycling_recovery_spin_01`

into `easy_aerobic`, while `cycling_recovery_spin_01` also maps to `recovery_or_rest`.

But these are intentionally different prescriptions:

- Recovery Spin: ~20–45 min, RPE 1–2, recovery objective;
- Zone 2: ~30–90 min, RPE 2–3, aerobic-volume objective.

The macrocycle requires **1–2 actual Zone-2 cycling exposures per week**. A short recovery spin should not be allowed to satisfy the aerobic-volume floor. The current `specificityCoverageContract.test.ts` only asserts that some day has the `easy_aerobic` key, so it can pass while the week contains no real Zone-2 ride.

**Recommendation:** either map `easy_aerobic` only to `cycling_zone2_standard_01`, or split the concepts into e.g. `aerobic_volume` vs `recovery_spin`. Add a hard regression that a week containing only recovery spins does **not** satisfy the Zone-2 minimum.

### 3. **Existing blocker remains valid: rest/recovery share spike is a failed acceptance condition**

The Phase 6.2c plan records healthy cycling scenarios moving from **14.3% → 50% rest/recovery**, loss of `strength_maintenance` in the stressed criterium scenario, and near-elimination of `modify` days. That is not a cosmetic baseline diff; §10 explicitly says not to bless the baseline when unexplained recovery/rest spikes occur.

Green CI therefore means implementation integrity is green, **not recommendation quality**.

The documented root cause is plausible: `Mobility/Recovery` has non-zero cost, yet often outranks zero-cost Rest even with `preferActiveRecovery=false`, which can keep fatigue oscillating above the recover threshold. I would test candidate policy fixes before choosing one, especially:

1. when `peakFatigue >= PROJECTED_FATIGUE_RECOVER_THRESHOLD`, prefer true Rest unless an active-recovery preference or specific objective justifies movement;
2. otherwise gate the Mobility/Recovery benefit advantage behind `preferActiveRecovery`;
3. only consider making Mobility/Recovery zero-cost if the catalog semantics genuinely support that.

Acceptance should include the fatigue actually clearing rather than only improving recommendation distribution.

### 4. **Upcoming travel is not yet a first-class plan block**

The active macrocycle has known travel **19–22 Aug** with a specific prescription: one 30–45 min hotel aerobic session, one 25–40 min hotel-gym maintenance session, walking counts as load, no maximal intervals.

`buildCyclingEventPlan()` deliberately does not create travel blocks; it relies on an explicit availability/day-context overlay or authored plan block. That is architecturally reasonable, but unless the production UI/data path actually supplies that date-range context, the recommender can still treat those dates as ordinary peak days and the `travel_aerobic`/`travel_strength` coverage roles are never activated.

**Recommendation:** before relying on the recommender for the 19–22 Aug period, either implement a real date-range travel overlay / authored block or add the trip explicitly as production data and prove with a scenario that normal quality work is suppressed while the two maintenance roles remain achievable.

### Suggested merge gate

I would require these contracts before updating the semantic baseline:

- [ ] A-priority Sep 5–6 branch preserves the 24–30 Aug race-specific peak week and only tapers from 31 Aug.
- [ ] Later Sep 12–13 confirmation gives one additional specific week and tapers from 7 Sep.
- [ ] Recovery Spin alone cannot satisfy the weekly Zone-2/aerobic-volume floor.
- [ ] Healthy high-fatigue scenarios recover below the recover threshold without multi-day Rest/Mobility lock-in.
- [ ] 19–22 Aug travel context produces one aerobic + one maintenance-strength opportunity, no maximal quality work.
- [ ] Re-run `simulate:diff`; only then use `simulate:update-baseline -- --reviewed` if the new distributions are coaching-plausible.

### Small maintenance note

The PR description's validation section still references head `ae16f0c4` / CI #180, while the current head is `6d11214` and CI #224 is successful. Worth refreshing before final merge so the review evidence matches the code being merged.

**Bottom line:** 6.2c is a meaningful architectural improvement, but the current policy still has a real semantic acceptance failure, and the generic A=14-day taper currently conflicts with the active race plan. I would fix those before blessing a new baseline or merging this as recommendation-policy complete.

Szczepanov:
## Handoff — macrocycle v5 recommendation-policy alignment

Current PR head: `3241f1b2a56f7e97cbad5042e34c47ff311649f6` on `codex/phase-0-5-completion-audit`.

### Landed commits
- `9c5cbc2` — **W0**: merge current `main` parity, including `ce43d13` (`droppedContributorObjectives` Firestore validation).
- `b706b71` — **W5 source authority (partial)**: add `docs/macrocycle-v5.md`, verified against the actual v5.0 macrocycle source. It confirms:
  - 24–30 Aug is the decisive peak week;
  - Branch A taper starts **31 Aug 2026**;
  - Branch B keeps 31 Aug–6 Sep specific and taper starts **7 Sep 2026**;
  - actual Zone-2 floor is **1–2 rides/week**;
  - 19–22 Aug travel prescription is one 30–45 min hotel aerobic + one 25–40 min maintenance-strength session, with no maximal intervals.
- `3241f1b` — **W3**: recovery-style authority + deterministic recover-tier Rest/Mobility ordering + modify-admissible full-body strength maintenance (`str_full_03`, systemic cost 0.45) + `fatigueClearing.test.ts`.

### Validation at this handoff
GitHub Actions **CI Pipeline #227** on `3241f1b` is fully green:
- Python Test Suite: success
- Docker Image Build: success
- frontend lint: success
- workout catalog validation: success
- policy-drift check: success
- frontend unit suite: success
- Firestore emulator suite: success
- simulation scenarios + aggregate bounds gate: success
- `simulate:diff`: success/advisory report completed
- production frontend build: success

### Important semantic caveat — do not treat green CI as blessing
During W3 staging, the old `cycling_criterium_A` assertion changed under rest-first recovery: rolling `race_specific_endurance` resolves **3/4 instead of 4/4**. The committed test now records 3/4 and explicitly says W1/W2 must replace that legacy aggregate-credit expectation with stronger date/role contracts. Revisit this before W6; do **not** infer recommendation quality from CI alone.

`docs/analysis/simulation-baseline.json` has **not** been blessed/updated. Keep it untouched until W6.

### Continue in this order
1. **W2 — aerobic coverage split**
   - `easy_aerobic` → required `aerobic_volume` + optional `recovery_spin`.
   - `aerobic_volume` only exact `cycling_zone2_standard_01`; recovery spin must not satisfy it.
   - add duration to `ExposureIdentity`; Zone-2 credit fails closed below workout minimum (20-min easy spins must not count).
   - strengthen `specificityCoverageContract.test.ts` to build a real `CoverageState`, resolve actual block phase, require authored target 2, and add recovery-spin-only + 20-min regressions.
   - wire `getUnfulfilledRequiredCoverage` / `getUnfulfilledTargetCoverage` into contracts/diagnostics.
   - run `npm run simulate:diff`; do not bless baseline.

2. **W1 — taper contract**
   - add `taperPolicy.ts` and remove duplicated 14/5/0 policy from `planSchedule.ts` + `periodization.ts`.
   - cycling A default = Monday-of-race-week (`week_aligned`, min 3 days); other cells preserve current 14/5/0 behavior.
   - add `EventTaperSpec` to `UserEvent`/`UserGoal`, validation, Firestore rules, goal→event mapping, Goals UI authoring.
   - hard `macrocycleContract.test.ts`: Sep 5 branch taper 2026-08-31; Sep 13 branch taper 2026-09-07; peak weeks retain sustained quality + outdoor event-specific; explicit start date overrides.
   - run `simulate:diff`; no baseline blessing.

3. **W4 — authored travel PlanBlock**
   - user-scoped `users/{uid}/plan_blocks`, service + validation + Firestore rules/emulator.
   - splice authored blocks into derived cycling plan; pass authored blocks through all planner/optimizer/sequence/trainingIntent/Home call sites.
   - 19–22 Aug travel block: exactly one `travel_aerobic`, one `travel_strength`, zero maximal-intensity picks.
   - documented seed script for the athlete’s real trip.
   - run `simulate:diff`; no baseline blessing.

4. **W5 remaining docs linkage**
   - link `docs/macrocycle-v5.md` from workout-library, ADR-0016 and Phase 6.2c plan/tests.

5. **W6 — observability + final acceptance**
   - ScenarioResult coverage summary + peak-fatigue trajectory/longest >0.65 run.
   - revisit rest ceiling (40.5%) after measured W3/W2/W1/W4 behavior.
   - `POLICY_VERSION = 2026-08-macrocycle-alignment-v1`; move weekly-coverage-v3 to historical.
   - ADR-0016 split documentation + ADR-0017 taper/authored blocks + Phase 6.2c §9/§10 status.
   - review final `simulate:diff`, explicitly account for pre-existing 11→12 scenario drift plus new travel scenario, then **once only** run `simulate:update-baseline -- --reviewed`.

### Merge-gate reminders
Final review still needs all six boxes from the macrocycle-alignment plan: both taper branches, real Zone-2 floor, fatigue clearing + no Rest/Mobility lock-in + strength maintenance restoration, travel prescription, and reviewed final semantic diff/baseline blessing.

PR #18, opened 2026-08-10
Plan Phase 7: training intent, capacity and first-class planning modes

Szczepanov:
## Summary

Documentation-only Phase 7 proposal, based on merged `main` after PR #17.

- **Phase 7A / ADR-0018:** repair the healthy/fresh cycling role-coverage failure with a bounded, stateful required-role reservation search while retaining the greedy production planner.
- **Phase 7B / ADR-0017:** derive evidence-backed adaptation dose first, combine real session/time/window capacity, then pack exact role occurrences. Planning mode and event-plan capability remain separate, so existing non-cycling events preserve their demand-derived path.
- Defines one owner for every persisted preference, Firestore/read-boundary validation and recovery, shared allocation-report identity/lifecycle, overlay application, and policy provenance.

No engine, service, rules, simulator baseline, or workout-catalog behavior changes land in this PR.

## Domain invariants

- Safety, readiness, injury, spacing, and Rest-first recover behavior remain authoritative.
- Exact authored coverage identity remains required; no role is inferred from title, category, modality, or stimulus overlap.
- Evidence-derived dose is not silently lowered to match declared session count; insufficiency is a typed shortfall.
- Existing cycling, running, triathlon, strength, and general event scenarios retain event-directed semantics.
- The September cycling coverage set remains byte-identical.
- State-search exhaustion is explicitly unresolved, never reported as a safety miss.

## Risks and reviewer guidance

- Review the proposed evidence authority/provenance classifications and the conservative `AthleteTrainingState.unknown` policy before accepting ADR-0017.
- Review the fixed seven-day / 14-occurrence / 1,024-node reservation budget and its p95/p99 performance gates before accepting ADR-0018.
- Phase 7B must preserve zero behavioral diff for existing event-directed fixtures (apart from the separately reviewed Phase 7A correction).
- This PR is architecture/plan-only; it intentionally does not introduce the future Firestore schema or runtime behavior.

## Validation

Executed successfully:

- Markdown local-link target validation across all edited documents.
- Markdown heading-uniqueness validation across all edited documents.
- `git diff --check`.

No runtime validation was needed because this PR changes documentation only.

## Screenshots or recordings

Not applicable: this PR contains no UI or visual artifact changes.


Szczepanov:
## Detailed architecture review — Phase 7A / 7B

I reviewed the current PR head `33b01a9` against the post-#17 `main` state (`34ddc30`), the merged PR #17 contracts, the current engine models/periodization/taper code, and the PR #17 semantic-baseline follow-up.

**Overall:** the direction is strong and the split into **7A (fix current event-directed weekly allocation)** and **7B (first-class Evergreen / training intent)** is the right decomposition. I would **not accept ADR-0017 / ADR-0018 as implementation authority yet**, though. There are four substantive design issues I would resolve in the docs first; two of them would otherwise create contradictions with the PR's own acceptance criteria.

### What I agree with

- **No fake event for Evergreen.** This is the right abstraction. Human intent/capacity/preferences should be persisted; physiological strategy should be derived.
- **D-CAP is correctly framed as product policy, not scientific law.** The 2/3/4/5/6-session table is a reasonable starting policy, and `3 strength + 3 endurance` should remain an example for a six-session balanced athlete rather than a universal prescription.
- **Auto / Adaptive Hybrid should be the default.** Users should not need periodization expertise during onboarding.
- **Phase 7A should fix allocation, not tune Rest percentage.** The PR #17 follow-up correctly established that the remaining healthy/fresh failure is sequence-level role allocation, while true recover-tier Rest-first behaviour should remain intact.
- **Exact programming-role identity remains the correct coverage authority.** Do not regress ADR-0016 into modality/category inference.
- The constraint that Phase 7B must not silently recalibrate the already-tested event-directed population is good.

---

## 1. BLOCKER — D-MODE conflates planning mode with rich-plan capability

ADR-0017 currently says `event_directed` is effective only when the profile selects it **and** the focus event is a currently supported `cycling_event`; running, triathlon, strength-meet and general events fall back to Evergreen until they get their own `PlanDefinition` contract.

I think this is the wrong abstraction boundary.

Today the engine already treats non-cycling events as real events through `evaluatePeriodizationPhase` + demand-derived objectives. Existing simulation coverage includes `running_marathon_A`, `triathlon_olympic_A`, `strength_meet_powerlifting_B`, etc. They may not have the rich authored coverage plan that cycling has, but they are still genuinely **event-directed** athletes.

As written, a profile-less runner with a marathon would switch from the current event-directed demand path to the new Evergreen default. That directly conflicts with Phase 7B's own acceptance condition:

> `simulate:diff` shows zero change on all pre-existing event-directed scenarios.

### Recommended model

Separate **user planning mode** from **event-plan implementation capability**:

```ts
PlanningMode = 'evergreen' | 'event_directed'

EventStrategy =
  | 'structured_plan'      // e.g. current cycling PlanDefinition
  | 'demand_derived'       // existing running/triathlon/strength fallback
```

Effective `event_directed` should mean: the athlete selected event-directed and there is an eligible focus event, regardless of whether that category has a rich coverage-set/PlanDefinition implementation.

Then:

- cycling event → `event_directed + structured_plan`
- running/triathlon/strength event → `event_directed + demand_derived` for now
- no eligible event → Evergreen
- explicit Evergreen profile → Evergreen even if an event exists, unless product UX explicitly changes the profile when the user chooses to target that event

This preserves existing semantics and makes future sport-specific plans an additive capability rather than a redefinition of what "event-directed" means.

I would change D-MODE, 7.2 and the compatibility tests accordingly.

---

## 2. BLOCKER — `TrainingIntentProfile` creates multiple authorities for the same preferences

The proposed profile currently contains:

- `defaultWeekdayMinutes`
- `defaultWeekendMinutes`
- `modalityPreferences`
- `conservativeBias`

But `UserPreferences` already owns:

- `defaultWeekdayTimeMin`
- `defaultWeekendTimeMin`
- `preferredModalities`
- `deprioritizedModalities`
- `avoidedModalities`
- `conservativeBias`
- `preferredRecoveryStyle`
- `extraRecoveryMargin`

The plan explicitly protects `preferredRecoveryStyle` as UserPreferences-only, but it does **not** define authority/migration/precedence for the other duplicated fields.

That gives us impossible-to-reason-about states such as:

```text
TrainingIntentProfile: Cycling = love
UserPreferences: Cycling = avoid
```

or profile `conservativeBias=false` while preferences say `true`.

### Recommendation

Make one ADR decision for **field ownership** before implementation.

My preference is:

**TrainingIntentProfile owns durable planning intent:**
- `planningMode`
- ordered training priorities
- weekly `min/target/maxSessions`
- future organization preference

**UserPreferences owns execution/user preferences:**
- weekday/weekend session duration
- time of day
- modality preference/exclusion
- recovery style
- conservative bias / recovery margin
- verbosity / units

If the richer `love/prefer/neutral/avoid/unavailable` modality scale is desired, evolve/migrate the existing UserPreferences modality representation rather than creating a second source of truth. `unavailable` can then feed the existing hard restriction path exactly as 7.7 proposes.

An alternative is to move these fields into `TrainingIntentProfile`, but then the plan needs an explicit one-time migration and must stop reading the old UserPreferences fields after migration. What I would avoid is merging two live authorities indefinitely.

---

## 3. BLOCKER — D-ORG should not persist unsupported modes that can break planning

The ADR proposes persisting:

```ts
'auto' | 'linear' | 'undulating' | 'block'
```

while only `auto` is implemented, and says the engine should fail loudly rather than silently ignore another stored value.

I agree that silently ignoring a user's choice is bad. But **persisting a valid value that makes recommendation generation fail is worse**.

A profile written today with `organizationPreference: 'block'` would be valid persisted data but intentionally unsupported runtime state.

### Recommendation

For this phase, either:

1. Persist only `organizationPreference: 'auto'` and widen the enum/schema when another organization is actually implemented; **preferred**, or
2. Persist a non-operative `requestedOrganization` separately for research/product intent, while `resolvedOrganization` remains `auto` and the UI clearly says the requested strategy is not yet active.

Do not make normal recommendation generation fail because the user selected a value the schema explicitly allowed.

The original research conclusion still holds: Linear / Undulating / Block are advanced choices; comparative evidence does not justify asking ordinary users to choose one on first run.

---

## 4. BLOCKER — Phase 7A's bipartite matching is not sufficient for stateful feasibility

ADR-0018 is directionally correct: reserve required roles before discretionary support work and keep the production greedy planner.

The technical issue is that the proposed **maximum-cardinality one-session-per-date matching** appears to treat feasibility edges as if they were independent. They are not.

Future-session feasibility depends on the state created by other future selections:

- fatigue changes after the first reserved session;
- hard-lower-body / quality spacing changes;
- anchor protection changes;
- projected history changes;
- a template that is individually feasible on Friday and another individually feasible on Sunday may not be jointly feasible when Friday's session is actually inserted.

Recalculating after every pick prevents unsafe execution, but it means the initial "maximum achievable matching cardinality" can be an overestimate. D-SUPPORT then cannot truthfully promise that a support candidate preserves the **maximum achievable** role cardinality if the matching model itself ignores pairwise/state-transition conflicts.

### Recommendation

Keep D-NO-BEAM, but make the required-role reservation solver **stateful**.

Because the horizon is only a rolling week and the solver considers only minimum required role occurrences (not every utility-ranked session sequence), a bounded deterministic DFS/backtracking/constraint search is small enough:

1. start from immutable today/tomorrow seeds;
2. choose the most constrained required occurrence;
3. try exact eligible date/template assignments in deterministic order;
4. apply the real projected state transition through the shared 7A.2 helper;
5. recurse on remaining required occurrences;
6. maximize fulfilled required occurrences, then deterministic tie-breaks;
7. only after those reservations are established does the existing greedy utility planner fill unreserved days.

That is **not** adoption of `sequenceSearch.ts` / beam search: it searches only the small set of must-have role reservations, not the whole workout-utility horizon.

If you want to keep simple bipartite matching, then weaken ADR-0018's guarantee language and treat it as a heuristic. I prefer the bounded stateful reservation solver because the entire point of 7A is to make the weekly role guarantee trustworthy.

Add a regression where two role/date edges are individually feasible from the seed state but the pair is jointly infeasible because of spacing/fatigue. That test will distinguish the two designs.

---

## 5. Strong recommendation — generalize the coverage naming, not just the registry

7.4/7.5 correctly recognize that coverage needs a registry, but the proposal still adds Evergreen by putting `'general'` into `EventPlanPhase` and reusing `EventPlanCoverageKey` / `EventPlanSessionCoverage`.

That leaves the supposedly first-class non-event architecture expressed through types literally called `EventPlan*`.

This is the best time to remove that conceptual coupling.

I would introduce generic names, for example:

```ts
PlanPhase
PlanCoverageKey
PlanSessionCoverage
CoverageSetDescriptor
```

and make the September cycling event set one descriptor using those types. Evergreen then becomes another peer descriptor instead of a special non-event value stuffed into `EventPlanPhase`.

This is mostly naming/type refactoring, but it matters because the whole architectural goal of this PR is to stop treating "no event" as an event-shaped edge case.

---

## 6. D-CAP needs explicit cardinality semantics for `min / target / max`

The table itself is good as initial policy, but the plan should define what each number means to the coverage/objective ledger.

For example with:

```text
min=2, target=3, max=4
```

I suggest the contract explicitly say:

- required/minimum role occurrences must be satisfiable within the athlete's minimum viable schedule;
- target coverage should fit `targetSessions`;
- optional/stretch work may use capacity up to `maxSessions`;
- one session may fulfil multiple adaptation objectives, but may fulfil multiple programming roles **only when exact coverage identity explicitly grants those keys**;
- derived distinct required occurrences must never silently exceed the capacity they are supposed to fit.

This is especially important for the proposed 2-session "multi-component" concept. The current catalog/coverage mechanics should not imply a combined strength+endurance session unless an authored workout actually represents that combined session. `1 strength + 1 endurance` is an honest two-session fallback; fabricated cross-role credit would violate ADR-0016.

Also, I would avoid `0` as a valid persisted default session duration when `minSessions >= 1`. If "unknown" is needed, make the duration optional/null; otherwise use a meaningful positive minimum.

---

## 7. Smaller documentation / sequencing fixes

### PR description is stale

#17 is now merged (`34ddc30`) and PR #18 is actually based on `main`. The current description still says it is stacked on #17 because `main` lacks the referenced docs. That is no longer true and should be refreshed.

### Source analysis is not in the repository

ADR-0017 and Phase 7B cite `training_intent_periodization_architecture_analysis.md`, but that file is not present at the repository root or under `docs/` on this head. Since the ADR explicitly uses it as source authority, please add the Markdown analysis under `docs/analysis/` (or replace the reference with a durable in-repo source).

Also update the "checked against PR #17 head `21a22a1`" / "verified against `6c79092`" references to the merged/current source state where appropriate. The core G1–G10 findings I spot-checked still exist on merged `main`, but the evidence record should point at the code readers will actually compare against.

### Taper defect

The `general_target` 5-star → A → 14-day taper defect is real on current `main`; I verified the category-agnostic legacy A/B fallback still exists. I agree with 7.7's fix. Since it is independent and behaviorally wrong today, I would consider landing that as a small isolated bugfix before the rest of 7B rather than waiting for the full Evergreen increment.

---

## Suggested acceptance decision

I would **approve the overall Phase 7 direction**, but revise the two ADRs before accepting them:

1. redefine D-MODE so event-directed is not cycling-only;
2. add explicit source-of-truth/migration ownership between `TrainingIntentProfile` and `UserPreferences`;
3. restrict D-ORG to implemented values (or separate requested vs resolved organization);
4. make 7A reservation feasibility stateful, or weaken the guarantee language;
5. preferably generalize `EventPlan*` coverage types before Evergreen becomes a peer planning mode;
6. clarify min/target/max capacity invariants;
7. add the missing source analysis and refresh the now-stale #17/base references.

After those changes, I think this is a strong blueprint. The important product principle is preserved: **the athlete supplies intent, commitment and preferences; the engine derives physiology and weekly structure; events are an overlay/authority when actually targeted, not a prerequisite for having a coherent training plan.**

Szczepanov (commented):
## Science-first policy review: make evidence the default authority, not a fixed session table

I would revise ADR-0017 before accepting it as implementation authority. The overall direction is right, but after checking the current evidence I think the hierarchy should be stronger and more explicit:

> **Science defines the safe/effective envelope and default prescription. Athlete goals + real capacity determine what can fit. Preferences choose among viable/equivalent options and may knowingly trade efficacy for adherence. Safety cannot be overridden.**

That is subtly different from both (a) “ask the athlete how many aerobic/strength sessions they want” and (b) “encode a 2/3/4/5/6-session table and call it science.”

### 1. Challenge: science usually gives ranges and outcome-specific dose variables, not a universal weekly session split

The strongest evidence does **not** support a universal rule like `6 sessions => 3 strength + 3 endurance`.

For general health, WHO gives a weekly *volume* floor: 150–300 min moderate aerobic activity (or 75–150 min vigorous equivalent) plus muscle strengthening on at least 2 days/week. That does not imply a particular number of standalone sessions, and mixed sessions can satisfy multiple requirements.

Source: WHO physical-activity guidance: https://www.who.int/europe/news-room/fact-sheets/item/physical-activity

For resistance training, the new 2026 ACSM position stand synthesized 137 reviews / >30,000 participants. The useful prescription variables are outcome-specific: strength is helped by heavier loads, full ROM, 2–3 sets and >=2 sessions/week; hypertrophy is helped by higher weekly volume (>=10 sets/week in the evidence synthesis); power favors moderate loads moved fast with low/moderate volume. Importantly, periodization itself did **not** consistently improve outcomes across the evidence base.

Source: ACSM 2026 Position Stand, PMID 41843416: https://pubmed.ncbi.nlm.nih.gov/41843416/

For endurance intensity distribution, there is also no single universal winner. A 2024 meta-analysis found a small VO2peak advantage for polarized training, especially in shorter interventions/highly trained athletes, but time-trial, TTE and threshold outcomes were similar. A newer individual-participant network meta-analysis found no overall POL-vs-PYR difference and suggested training status may moderate the response (competitive athletes may lean POL, recreational athletes may lean PYR).

Sources: https://pubmed.ncbi.nlm.nih.gov/38717713/ and https://pubmed.ncbi.nlm.nih.gov/39888556/

So I recommend treating the current D-CAP `2→..., 3→..., 4→2+2, 5→..., 6→3+3` table as a **fallback heuristic only**, not the primary prescription model.

### 2. Better abstraction: prescribe adaptation dose first, then pack it into sessions

The domain model should distinguish **what physiology we want** from **how many calendar sessions deliver it**.

Conceptually:

```text
athlete goal / event
        +
training status
        +
evidence-backed dose policy
        ↓
required adaptation envelope
(aerobic volume, quality exposure, strength volume/frequency,
power/speed, skill/specificity, recovery)
        +
weekly capacity / available days / time
        +
modality and environment constraints/preferences
        ↓
weekly allocation / session packing
        ↓
readiness-guided daily execution
```

This fits ADR-0016 particularly well because one workout may legitimately provide multiple *adaptation credits* while still fulfilling only exact authored *coverage roles*.

A football match, CrossFit session, ski session, long ride or mixed gym/cardio session should not be forced into one “strength” or “endurance” bucket if its actual stimulus profile satisfies several adaptations.

### 3. Replace D-CAP's session-allocation authority with an evidence-derived weekly prescription

I would change the wording from:

> derived objective counts and coverage minimums are a pure function of capacity plus stated priorities under a versioned policy table

into something closer to:

> **Goal/event + training status select an evidence-backed dose policy. Weekly capacity constrains and packs that dose into feasible sessions. Preferences select among admissible implementations. When capacity cannot satisfy the full dose, the engine preserves the highest-value minimum effective coverage and exposes the shortfall.**

`weeklyCommitment` is still valuable, but `targetSessions` cannot be the main physiological input. Three 25-minute sessions and three 90-minute sessions are not the same capacity.

I would retain at least:

```ts
weeklyCommitment: {
  minSessions: number;
  targetSessions: number;
  maxSessions: number;
  defaultWeekdayMinutes: number;
  defaultWeekendMinutes: number;
}
```

but derive a **weekly time budget** and eventually day-level availability from it/fixed activities. The evidence policy should reason in the native dose variable where possible:

- aerobic: weekly minutes + intensity distribution / quality exposure;
- strength/hypertrophy: weekly hard-set volume / muscle-group frequency / load intent;
- power: quality exposures + low-fatigue explosive dose;
- sport specificity: event/sport exposure requirements;
- recovery: constraint resulting from load/readiness, not a quota to fill.

Sessions are then the packing container.

### 4. Introduce evidence strength explicitly; do not hide product heuristics as physiology

A major product risk is false precision. Exercise-science evidence is population-level, often heterogeneous, and much less certain for exact weekly architecture than for broad training effects.

I would make policy provenance first-class, even if initially only in code/docs:

```ts
type EvidenceAuthority =
  | 'guideline_floor'       // strong population recommendation
  | 'outcome_supported'     // strong direction, configurable dose range
  | 'conditional'           // depends materially on training status/context
  | 'product_heuristic';    // reasonable but not evidence-derived
```

Every important policy constant/table should say which of those it is and cite its source/ADR. This would make recalibration much safer than having all constants look equally authoritative.

Examples:

- WHO aerobic + 2-day strengthening floor for a health-priority adult → `guideline_floor`.
- >=2 resistance exposures/week and higher volume for hypertrophy → `outcome_supported`.
- POL vs PYR selection → `conditional`, not hard-coded truth.
- `targetSessions=5 => 2+3` → `product_heuristic` unless a more specific goal policy derives it.

### 5. Preferences matter scientifically too — but should usually select the implementation, not erase the adaptation

I agree strongly with giving the athlete preference authority. Adherence is part of realized effectiveness. Reviews of exercise motivation/adherence consistently support autonomy, enjoyment, flexibility and integrating exercise into daily life as important determinants of persistence.

Examples: https://pubmed.ncbi.nlm.nih.gov/22726453/ and https://pubmed.ncbi.nlm.nih.gov/39557793/

But preference needs a precedence contract:

| User statement | Engine meaning |
|---|---|
| `love` / `prefer` | choose this modality first when it can deliver the required adaptation |
| `neutral` | normal candidate |
| `avoid` | avoid if an evidence-equivalent substitute exists; otherwise expose the tradeoff |
| `unavailable` | hard feasibility exclusion |
| explicit informed override | user accepts a known goal-performance compromise; never bypasses safety |

Example: someone training for health who hates running can cycle, row, ski, swim, etc. There is usually no reason to force running.

But someone targeting a running race who says running is unavailable has a **goal/constraint conflict**, not an excuse for the engine to pretend cycling is fully race-specific. The correct UX is: preserve safe aerobic development, explain that race specificity cannot be fulfilled, and let the user change either the event goal or constraint.

That is better than either blindly overriding preference or blindly overriding science.

### 6. Define the actual authority order

I recommend ADR-0017 state this explicitly:

1. **Safety / medical / injury / hard recovery constraints** — absolute.
2. **Hard feasibility** — equipment, unavailable modality, available time/day.
3. **Event/goal specificity** — what adaptation matters.
4. **Evidence-backed dose envelope** — minimum/range/ceiling appropriate to goal and training status.
5. **Athlete weekly commitment** — determines what subset can realistically be delivered.
6. **Athlete preference** — chooses modalities, environment, schedule and equivalent session forms; may knowingly accept a performance tradeoff.
7. **Adaptive response/history/readiness** — modifies day-to-day dose and eventually personalizes the prior as individual response data accumulate.
8. **Utility/ranking heuristics** — only after the above.

I would summarize it as:

> **Science supplies the prior; the athlete supplies constraints and values; longitudinal response supplies personalization.**

### 7. D-MODE should remain independent from whether a rich sport-specific PlanDefinition exists

This reinforces my earlier review comment: `event_directed` should mean the athlete is actually targeting a real event, not “the event happens to be cycling and has the richest implementation.” A marathon remains event-directed even if the running event-plan implementation is incomplete.

The engine can separately expose `planCapability = rich | generic | partial` rather than silently reclassifying the athlete as Evergreen. Otherwise “science-driven” becomes implementation-driven.

### 8. D-ORG: I would ship only `auto` for now

The 2026 ACSM synthesis reports that periodization did not consistently affect resistance-training outcomes, and the older direct LP-vs-UP literature similarly does not establish a universal winner. Endurance POL/PYR evidence is also conditional rather than absolute.

Therefore I would **not persist user-selectable `linear | undulating | block` yet**. Ship `auto` only. Later, when an organization has an actual implementation and acceptance tests, add it as an advanced preference. The app can explain which structure Auto selected and why.

This keeps “show different approaches” as education/explainability without pretending the engine already has evidence that a novice should choose one.

### 9. Phase 7A allocation still belongs before 7B

I continue to support the 7A/7B split. The PR #17 follow-up correctly identified a sequence/allocation defect: healthy athletes can miss required roles despite safe capacity. Fixing that first gives the future science-derived Evergreen coverage something reliable to allocate into.

However, 7A should reserve **evidence/authored minimum requirements**, not arbitrary target counts. Safety-forced misses remain allowed and typed.

### 10. Concrete changes I recommend before accepting ADR-0017

1. Replace D-CAP's 2/3/4/5/6 table as the primary authority with **goal-specific evidence policy → weekly dose → session packing**.
2. Keep that table only as a documented fallback/product heuristic for sparse profiles.
3. Add an explicit evidence-confidence/provenance classification for policy constants.
4. Define the preference precedence contract above, including goal/preference conflicts and informed tradeoffs.
5. Do not duplicate preference ownership between `TrainingIntentProfile` and `UserPreferences`; one persisted authority per field.
6. Keep `event_directed` semantic even for sports with partial plan support; separate planning mode from engine capability.
7. Persist only `organizationPreference: 'auto'` until alternatives are implemented.
8. Add acceptance tests for:
   - same goal/capacity, different preferred modality => same required adaptation envelope where substitutes exist;
   - insufficient time => explicit minimum-dose shortfall rather than silently dropping requirements;
   - `unavailable` modality => never selected;
   - `avoid` modality => substituted when equivalent, conflict reported when specificity requires it;
   - health-priority profile => evidence floor represented in minutes + strength days rather than arbitrary session count;
   - strength/hypertrophy priority => dose changes by outcome, not merely “more Strength sessions”;
   - different training status can choose different endurance organization without claiming one universal TID;
   - safety/readiness can always reduce training below an evidence target without being treated as a planner failure.
9. Put the source analysis in `docs/analysis/` and add a small evidence table linking each initial policy rule to source, population, outcome and confidence.

### Recommended product principle

I would adopt this as the Phase 7 north star:

> **The user should not have to know exercise physiology. They state what they are training for, how much time they can invest, what they can/cannot do, and what they enjoy. The engine derives the best-supported training dose from current evidence, packs it into that real-world capacity, adapts it using readiness and observed response, and explains whenever preference or capacity means accepting a tradeoff.**

That gets much closer to a genuinely science-driven recommender than either a fixed “3 strength + 3 endurance” template or a user-configured training-plan builder.

Szczepanov (commented):
## Follow-up review after `c06c536` — much stronger; one substantive science-first gap remains

This revision addresses essentially all of the structural concerns from my previous review. In particular, I think the following changes are correct and should stay:

- **D-MODE now separates athlete intent from engine capability.** Running/triathlon/strength/general events remain `event_directed` and retain their existing demand-derived path; `structured_plan` is a capability, not the definition of an event-directed athlete. This fixes the biggest semantic problem in the earlier draft.
- **D-OWNERSHIP is now clean.** `TrainingIntentProfile` owns durable planning intent/capacity; `UserPreferences` remains the only owner of duration, modality, recovery style, conservative bias, etc. No competing preference models.
- **D-ORG now persists only `auto`.** Good. Do not persist values the engine cannot execute.
- **The Plan* vocabulary generalisation is the right direction.** Evergreen should not be forced through an event-shaped type system.
- **7A's allocator is now stateful.** The DFS/backtracking required-role search applies projected state transitions while searching, so two assignments that are independently feasible but jointly infeasible cannot both be reserved. This is materially better than the previous static bipartite-matching proposal and is still narrower than adopting the ADR-0015 whole-horizon beam search.
- **The architecture analysis is now in-repo and evidence/provenance is explicit.** Good improvement over an external/untracked source.
- **The PR description is now accurate**, and CI #240 is green.

### Remaining issue: the architecture still is not quite "science -> dose -> packing -> preference"

The stated product direction is to drive training by science as far as the evidence supports, and let athlete preference choose/override implementation where appropriate. The current Phase 7B design is closer, but `7.3` still makes this the core pipeline:

`targetSessions + first priority -> 2/3/4/5/6 allocation table -> WeeklyBudget -> objectives/coverage`

with evidence/provenance attached to the allocation constants.

That is still fundamentally a **session-allocation heuristic with citations**, rather than an evidence-derived training prescription that is subsequently packed into the athlete's real capacity.

I would change the conceptual order before accepting ADR-0017:

`goal / event demands`
`+ inferred athlete/training state`
`+ evidence policy`
`-> adaptation/dose requirements`
`-> capacity/availability feasibility`
`-> preference-based implementation choice`
`-> weekly role allocation`
`-> readiness/history adaptation`

The distinction matters because exercise science generally speaks in **dose dimensions**, not app-session counts: minutes of aerobic work, frequency/volume of resistance training, high-intensity exposure, progression, specificity, etc. A session is merely a container that can deliver several adaptations.

### 1. Add an explicit evidence-policy / dose layer before `WeeklyBudget`

I recommend introducing a pure concept along these lines (names illustrative):

```ts
interface AdaptationDoseRequirement {
  adaptation: AdaptationKey;
  minimum: DoseTarget | null;
  target: DoseRange;
  priority: 'required' | 'target' | 'optional';
  evidence: EvidenceProvenance;
  substitutionPolicy: SubstitutionPolicy;
}

interface EvidenceBackedStrategy {
  requirements: AdaptationDoseRequirement[];
  hardSessionCap?: number;
  warnings: PolicyWarning[];
}

resolveEvidenceBackedStrategy(goal, athleteState, eventContext): EvidenceBackedStrategy
```

Then `resolveWeeklyBudget`/packing takes that strategy plus real capacity and determines how many sessions/roles are needed to deliver it.

The existing 2→6 table can still exist, but it should be explicitly a **packing fallback/tie-breaker**, not the source of physiological requirements.

### 2. Weekly capacity cannot be represented by session cardinality alone

This is the clearest practical gap in the current proposed API:

```ts
resolveWeeklyBudget(commitment, priorities)
```

Only receives `minSessions/targetSessions/maxSessions`.

But the evidence table itself cites the WHO health floor in **minutes**. Three 25-minute sessions and three 90-minute sessions are radically different capacity. Likewise, a strength prescription cannot be assessed from session count alone if one athlete has 25 minutes and another 75.

It is correct that duration remains owned by `UserPreferences`; ownership and consumption are separate questions. The planner should build a resolved capacity object from both authorities, e.g.:

```ts
interface ResolvedTrainingCapacity {
  minSessions: number;
  targetSessions: number;
  maxSessions: number;
  weekdayMinutes: number;
  weekendMinutes: number;
  availableWindows?: ...;
  estimatedTargetWeeklyMinutes: number;
}
```

`TrainingIntentProfile` need not duplicate duration, but the science/dose resolver and packer must see it.

Otherwise Phase 7 cannot actually evaluate the strongest evidence floor it documents.

### 3. Training status/history must influence the science policy

I do not see athlete training status in the proposed `resolveWeeklyBudget(commitment, priorities)` inputs. That will eventually make two athletes with the same declared priority/capacity receive essentially the same starting prescription even if one is untrained and the other has years of high-volume training.

Do **not** make the user answer a physiology questionnaire if the app can infer this. Derive a coarse `AthleteTrainingState` from existing history/Garmin/completed exposure where possible, with a conservative unknown fallback. Examples of useful policy inputs:

- recent weekly frequency and duration;
- recent strength exposure;
- recent aerobic/quality exposure;
- training consistency / training age proxy;
- current tolerated load and progression trend;
- sport-specific history when relevant.

The scientific prior should depend on that resolved state; readiness then modifies execution day-to-day.

### 4. Make the science-vs-preference conflict contract explicit

D-OWNERSHIP now says where preferences live, but I would add one policy decision explaining **what they are allowed to do**.

Suggested hierarchy:

1. safety/injury/hard medical/feasibility constraints;
2. goal/event specificity;
3. evidence-backed minimum/target dose envelope;
4. real capacity;
5. athlete preference among physiologically valid implementations;
6. readiness/history modifies execution;
7. product ranking heuristics last.

Preference should normally choose *how* an adaptation is delivered, not redefine the adaptation itself.

Examples:

- General-health athlete hates running -> cycling/rowing/etc. can substitute if they satisfy the same adaptation requirement. No problem.
- Marathon athlete marks running `unavailable` -> do **not** silently treat cycling as equivalent marathon-specific preparation. Preserve transferable aerobic work, emit a typed `goal_constraint_conflict` / specificity shortfall, and explain the expected trade-off.
- Athlete prefers 3 days when the evidence-backed minimum cannot fit -> do not mutate the scientific minimum to 3 days. Return a transparent `minimum_dose_shortfall` and produce the best feasible plan.

This is the concrete meaning of “science first, preference override where appropriate.” Safety should never be preference-overridable, while preferences may knowingly trade expected efficacy for adherence when the system can explain the consequence.

### 5. Tighten D-CAP wording so `minSessions` does not become the definition of scientific minimum

Current wording says all required roles must fit inside `minSessions`, otherwise emit a shortfall. The shortfall behavior is correct, but be careful about derivation order.

I would explicitly say:

> Evidence/goal policy derives the minimum effective dose independently of declared capacity. `minSessions` is the athlete's guaranteed packing capacity, not the physiological definition of the minimum. When the dose cannot be represented inside that capacity, the engine reports `minimum_dose_shortfall` and packs the safest/highest-value feasible subset without fabricating credit.

That protects the science-first architecture from slowly turning every evidence floor into “whatever fits the user's minimum session count.”

### 6. Evidence provenance is a strong start; make it executable policy metadata

The new analysis table is exactly the right direction. I would make the contract slightly stronger before implementation: every policy constant that claims scientific authority should carry at least:

- source / citation id;
- population;
- outcome supported;
- evidence strength/confidence;
- applicability conditions;
- whether it is a `guideline_floor`, `outcome_supported_default`, `conditional_prior`, or `product_heuristic`;
- policy/version/date reviewed.

This prevents a WHO population-health floor, an ACSM outcome-specific recommendation, and the repository's 5-session `2+3` packing heuristic from appearing equally authoritative in code.

I would also avoid labeling the entire Evergreen coverage set “evidence-backed.” Some of its roles will initially be evidence-derived; others will still be product policy. The provenance needs to live at the rule/requirement level.

## Suggested Phase 7B sequence

I would slightly reshape 7.3 rather than expanding the whole phase:

**7.3a — Resolve evidence-backed adaptation/dose requirements**
- input: planning goal/event context + inferred athlete training state;
- output: minimum/target/optional adaptation dose requirements with provenance;
- no session packing yet.

**7.3b — Resolve real training capacity**
- combine `TrainingIntentProfile.weeklyCommitment` with duration/availability from the existing owners;
- output a capacity object in sessions + time/windows.

**7.3c — Pack dose into sessions/roles**
- use exact workout/coverage identities;
- use the 2–6 table only as a low-confidence packing heuristic when multiple valid packings remain;
- emit typed shortfalls when science/goal requirements exceed capacity.

Then 7.4–7.9 can remain largely as currently written.

## Verdict

**Very strong revision. ADR-0018 / Phase 7A now looks conceptually ready for acceptance.** I would not reopen its safety/recovery direction based on this review.

**ADR-0017 / Phase 7B is close, but I recommend one more architecture pass before accepting it:** make evidence-derived adaptation/dose requirements first-class and make session/time capacity a downstream packing constraint. The current provenance table is valuable, but if `targetSessions + priority` remains the thing that creates the training strategy, the implementation will still be product-policy-first rather than science-first.

Once that ordering is explicit, I think the Phase 7 architecture will be on a very solid foundation.

Szczepanov:
Implemented the latest architecture-review follow-up in `fc55488` (**Refine Phase 7 dose and allocation contracts**).

### Phase 7B / ADR-0017
- Reordered Evergreen planning to **evidence-backed adaptation dose → resolved time/session capacity → exact weekly role packing → day-to-day readiness adaptation**.
- Added conservative inferred `AthleteTrainingState`, per-requirement evidence provenance/authority classes, and explicit science-versus-preference/substitution rules.
- Made minutes and availability windows first-class capacity inputs without duplicating `UserPreferences` ownership.
- Defined zero/missing-duration handling, typed `minimum_dose_shortfall` / `goal_constraint_conflict`, Firestore enum/field validation, and legacy read recovery diagnostics.
- Defined a shared overlay step so authored travel blocks and fixed activities constrain structured, demand-derived, and Evergreen paths.

### Phase 7A / ADR-0018
- Defined canonical required-role occurrence identity, shared allocation-report fields/lifecycle, typed miss reasons, and `wasMoved` as an annotation rather than a terminal status.
- Added deterministic reservation/viability limits: seven dates, 14 occurrences, four exact candidates per occurrence, 1,024 state-transition nodes, with a deterministic best-known partial result and `unresolved_search_budget` on exhaustion.
- Clarified that discretionary Rest participates in viability protection, while true recover-tier Rest remains safety-first and is never overridden by reservation demand.

Also refreshed the PR description with reviewer guidance and validation results.

Validation: local Markdown-link target check, heading-uniqueness check, and `git diff --check` passed. This commit remains documentation-only; no runtime behavior changed.

Szczepanov (commented):
Re-review at `fc554884` after the latest architecture revision.

**Verdict: conceptually approved as the Phase 7 architecture/plan proposal.** The previous substantive concern is addressed. The design now has the ordering I wanted rather than attaching evidence labels to a session-count heuristic:

`goal/event + inferred athlete state + evidence policy -> adaptation-dose requirements -> real session/time/window capacity -> preference-compatible implementation -> exact role packing -> readiness/history execution adaptation`.

Specific improvements I verified:

- **D-DOSE is now a real upstream policy layer.** `resolveEvidenceBackedStrategy` derives `AdaptationDoseRequirement`s independently of session count, with per-requirement provenance and substitution policy.
- **D-CAP no longer defines physiology.** `ResolvedTrainingCapacity` combines session cardinality with weekday/weekend duration and actual usable windows; insufficient capacity produces a typed shortfall rather than silently lowering the dose.
- The **2–6 session table is correctly demoted** to a low-confidence packing fallback/tie-breaker after dose, capacity and eligible implementations are known.
- **Preference semantics are correct:** preference selects among permitted substitutions; it cannot fabricate specificity or coverage. The marathon-running-unavailable example is exactly the right conflict contract.
- `AthleteTrainingState` is separated from daily readiness and has an explicit conservative `unknown` path instead of pretending training status is known.
- Preference ownership/read-boundary recovery is now explicit enough to avoid two live sources of truth.
- ADR-0018 is substantially stronger: stateful joint feasibility, deterministic search budget, explicit `unresolved_search_budget`, no wall-clock semantic cutoff, shared allocation-report lifecycle, and discretionary Rest gets the same viability protection while true recover-tier Rest remains authoritative.
- The fixed Phase 7 baseline and zero-diff compatibility contract for existing event-directed paths remain intact.

### One science-semantics refinement I would make before marking ADR-0017 `Accepted`

Be careful with the word **minimum** for population guidelines. The WHO 150–300 min/week recommendation is a high-confidence public-health target/range, but 150 min is not a biological threshold below which benefit disappears. There is meaningful dose-response below the guideline target. If a health-priority athlete can currently support only 120 min/week, the engine should not imply that 120 min is physiologically worthless or that it failed a true minimum-effective-dose requirement.

I suggest distinguishing at least:

- `guideline_target_shortfall` / `below_guideline_range` for population recommendations such as WHO;
- `goal_requirement_shortfall` for specificity/performance requirements whose absence materially prevents the stated goal;
- `minimum_dose_shortfall` only where the policy really has evidence for a minimum admissible/effective dose, rather than simply the lower edge of a recommended range.

Equivalently, `AdaptationDoseRequirement.minimum` could carry semantics/authority (`guideline_recommended_minimum`, `goal_required_minimum`, `evidence_supported_minimum`) instead of every lower bound flowing into the same generic `minimum_dose_shortfall`.

This matters because the new architecture is explicitly science-first: it should preserve uncertainty and dose-response rather than convert a guideline boundary into false biological precision.

A smaller related implementation note: give `AthleteTrainingState` an inference confidence/data-quality field (or equivalent diagnostics). Recent Garmin/history can support `recentTrainingState`, but it cannot reliably establish literal training age. Conditional priors should be allowed to fall back toward `unknown` when history is sparse or contradictory.

Neither point changes the overall architecture or reopens the previous review. With those semantics tightened before implementation/ADR acceptance, I think Phase 7B is on a strong foundation. Phase 7A/ADR-0018 is conceptually ready to implement under its stated acceptance/performance gates.

PR #31, opened 2026-08-12
🧹 [Refactor] Extract complex loop logic in planner

Szczepanov:
🎯 **What:** The code health issue addressed
Extracted the heavily nested and complex `for` loop body in `app/src/engine/planner.ts:1188` into a nested helper function `evaluateForecastDate(offset: number)`.

💡 **Why:** How this improves maintainability
The inner loop contained roughly 180 lines of logic involving various variable updates and side effects, making it hard to follow. Extracting it clarifies the logic of what happens during each step of the forecasting without breaking the planner, as all closure scope variables remain accessible exactly as they were.

✅ **Verification:** How you confirmed the change is safe
Extracted the function to the immediate scope outside the for loop, verified no `continue`, `break`, or early `return` issues would arise, and successfully passed `npm run check` which covers linting, typechecking, and tests.

✨ **Result:** The improvement achieved
The for loop itself now acts cleanly as a driver that iteratively invokes `evaluateForecastDate(offset)`, vastly improving code readability.

---
*PR created automatically by Jules for task [13731167211251075760](https://jules.google.com/task/13731167211251075760) started by @Szczepanov*

### DIFF
--- a/app/src/engine/planner.ts
+++ b/app/src/engine/planner.ts
@@ -1185,7 +1185,7 @@ export function generateWeekAheadPlan(
     const settledOutcomes = new Map<string, WeeklyRoleAllocationOutcome>();
     const displacementReasons = new Map<string, WeeklyRoleMissReason>();
 
-    for (let offset = resultDays.length + 1; offset <= totalDays; offset++) {
+    const evaluateForecastDate = (offset: number) => {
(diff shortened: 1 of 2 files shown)


Szczepanov:
@jules sync main and fix ci

PR #45, opened 2026-08-15
Freeze September cycling coverage contract

### LINKED ISSUE #43: No frozen-reference guard for `SEPTEMBER_CYCLING_EVENT_SESSION_COVERAGE`
## Problem

`SEPTEMBER_CYCLING_EVENT_SESSION_COVERAGE` (`workouts/event-plan.ts`) is described across the docs as the **frozen, event-directed cycling contract** — see `docs/architecture/recommendation-engine.md` § *Planning authority and coverage sets* and ADR-0016.

Its preservation is not mechanically proved. The only test references are in `engine/microcycle.test.ts`, where it is passed as *input* to `buildPlanDefinition` — that exercises the builder, not the constant's contents. An edit to the coverage set itself would pass the suite.

This was finding 3 of [`docs/analysis/2026-08-11-phase-6-7-compliance-review.md`](docs/analysis/2026-08-11-phase-6-7-compliance-review.md), recorded as "supported by current source shape, but not mechanically proved against a frozen reference."

## Why it matters

"Frozen" is a claim the docs make and the suite does not back. A silent edit to a coverage key, phase list, or minimum-session count would change event-directed planning for the September event with no test failure and no `POLICY_VERSION` signal — `check-policy-drift.mjs` watches decision logic, not this data.

## Suggested fix

A byte-equality or structural snapshot assertion against a committed reference, in a test that states in its failure message that changing the constant is a deliberate contract change requiring an ADR amendment — not a test to update reflexively.

Note the tension with `simulate:diff`'s design rationale (**D-GATE**: a byte-exact gate on *decision output* would freeze known-bad behaviour). That argument does not apply here: this is a frozen *input contract* the docs already claim is immutable, not an output whose improvement should be allowed to change it.

Szczepanov:
## Summary

- add a direct frozen-reference regression for `SEPTEMBER_CYCLING_EVENT_SESSION_COVERAGE`
- hash the complete ordered JSON serialization so changes to keys, phases, requirements, workout mappings, labels, notes, or ordering fail the contract test
- make the failure message explicit that a deliberate change requires an ADR/macrocycle contract amendment rather than routine snapshot refresh

## Validation

- expected digest was computed from the current 18-entry frozen contract on `main`
- CI is the authoritative validation for this connector-authored branch

Closes #43

### DIFF
--- a/app/src/workouts/event-plan-contract.test.ts
+++ b/app/src/workouts/event-plan-contract.test.ts
@@ -1,0 +1,25 @@
+import { createHash } from 'node:crypto';
+import { describe, expect, it } from 'vitest';
+import { SEPTEMBER_CYCLING_EVENT_SESSION_COVERAGE } from './event-plan';
+
+/**
+ * Frozen serialization of the ADR-0016 / macrocycle-v5 September cycling coverage
+ * contract as it existed when the generic coverage registry was introduced.
+ *
+ * Do not update this digest as routine snapshot maintenance. A deliberate change to the
+ * authored event-directed contract requires an explicit ADR/contract amendment and review.
+ */
+const SEPTEMBER_CYCLING_EVENT_COVERAGE_SHA256 = 'cf5d082175c20ae831db0f753a911b2e70661cbc0a5950cac90597e571be8852';
+
+describe('September cycling event coverage contract', () => {
+    it('remains byte-for-byte frozen unless the governing ADR/contract is deliberately amended', () => {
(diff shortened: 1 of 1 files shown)


Szczepanov:
Closing without merging. We're choosing not to add a frozen-reference regression guard for SEPTEMBER_CYCLING_EVENT_SESSION_COVERAGE at this time; closing #43 alongside this for the same reason.

PR #50, opened 2026-08-16
Phase 8: externally-planned mode — import, adjudicate, critique, replay

Szczepanov:
## Summary

- Completes **Phase 8** ([plan](docs/plans/phase-8-externally-planned-mode.md)), implementing [ADR-0019](docs/adr/0019-externally-authored-plans-and-session-adjudication.md): a third planning mode, `externally_planned`, in which a plan authored outside this app owns *selection* and the engine keeps *adjudication* — safety, feasibility, readiness ceilings, dose, weekly critique, and replay.
- **Why now:** the athlete already reassesses training with a general-purpose AI every week or two and prefers its macro/meso structure to the 27-entry catalog this engine ranks. Until now the only way to use both was to read the plan on a phone and ignore the app's daily signal.
- **8.7 weekly critique** — `engine/externalCritique.ts` reviews the placed week using the selection modules the mode no longer drives (the microcycle objective ledger, `planner.ts`'s forward fatigue projection, `optimizer.ts`'s recovery-spacing constraints, and the athlete's declared weekly commitment). Advisory only: the module cannot import `externalSession.ts`, and an architecture test enforces that boundary in both directions.
- **8.8 provenance and replay** — `RecommendationAudit` gains `externalPlan { planId, revision, sessionId, contentHash }`. `replay.ts` verifies an external decision against the revision it names, and a plan re-imported under the same revision number fails with an explicit hash-mismatch reason rather than replaying as if unchanged.
- **8.9 UI** — an import screen (paste → validate → preview with a per-session diff on re-import → confirm, then an invitation to confirm derived objectives), a Home verdict banner that leads with the decision and names the failing gate in plain language, and a week view carrying the critique findings plus the missed-session proposal with accept / choose another day / drop.
- **User impact:** additive. An athlete on `evergreen` or `event_directed` sees no change — `simulate:diff` reports no modified baseline scenario. Setting `planningMode` back restores previous behaviour completely; stored plans become inert data.

## Validation

- [ ] `uv run pytest` (backend changes) — not applicable, no Python changed.
- [ ] `uv run ruff check .` and `uv run mypy src/garmin_sync` (backend changes) — not applicable, no Python changed.
- [x] `cd app && npm run check` (frontend changes)
- [x] `cd app && npm run test:rules` (Firestore rules changes)
- [x] `cd app && npm run simulate:scenarios` (recommendation-engine changes)
- [x] `cd app && npm run visual:refresh` (material UI changes)

Results:

- `npm run check`: **pass** — typecheck, ESLint, 934 tests across 78 files, workout catalog validation.
- `npm run test:rules`: **pass** — 40 tests in the Firestore emulator, including a new one covering the `externalPlan` audit field. Reverting just the `hasOnly` change in `firestore.rules` fails that one test and nothing else, so the guard is real rather than inert.
- `npm run simulate:scenarios`: **pass** — 24 scenarios simulated, committed baseline unchanged.
- `npm run simulate:diff`: **pass** — zero `[MODIFIED]` and zero `[REMOVED SCENARIO]` lines. It also prints 11 `[NEW SCENARIO]` lines; those predate this branch (the committed baseline was never refreshed after Phase 7 added scenarios) and are identical before and after these commits.
- `node scripts/check-policy-drift.mjs <base>`: **pass** — "3 engine file(s) modified and POLICY_VERSION was correctly updated."
- `npm run visual:refresh`: **pass** — 24 captures. A new `home-externally-planned` fixture was added so the harness actually exercises the new screens rather than passing without covering them.
- Manual check: reviewing the captured `home-externally-planned` screenshot showed the verdict rationale printing the raw enum `(time_limit)` to the athlete. Fixed at the source — `describeEligibilityReasons` now lives beside the reasons in `eligibility.ts` so the persisted rationale and the rendered banner cannot explain the same exclusion two different ways — with a test asserting no gate code reaches the prose.
- `node scripts/check-firestore-rules-drift.mjs`: **not run** — it requires an explicit live Firebase project (`--project`), which this environment has no credentials for.

## Risk and reviewer guidance

**Look hardest at `app/firestore.rules`.** `hasValidRecommendationAudit`'s `hasOnly` list is a closed set, and the first draft of this branch omitted `externalPlan` from it. That denied every externally-planned day's write outright; because `Home` only `console.warn`s on a failed save, the entire mode would have silently persisted nothing — no adherence prompt, no `yesterdayRec`, no history — while appearing to work. It is fixed and covered by an emulator test, but the same class of failure is one field away for any future audit change.

Second, the **adjudication/selection boundary**. `externalSession.ts` must never become a second ranking path and `externalCritique.ts` must never be able to change a verdict (D-CRITIQUE). Both are enforced structurally by `externalArchitecture.test.ts`, which walks the runtime import graph (type-only imports excluded, since they are erased) and includes a positive control so a `null` result means absence rather than a broken scan.

Third, the **`scale` verdict's dose**. `resolveExecutionDose` caps by readiness *tier*, which is derived independently of `mode`, so a `scale` triggered only by the `modify` cost ceiling previously returned the full planned volume — "do the reduced version" printed over the full prescription. It now cuts by the smaller of the author's own `reducedDurationMin` ratio and proportional relief against the ceiling. That proportional fallback is the one number in this branch the engine invents rather than reads; it assumes cost is roughly linear in duration at fixed intensity.

**Compatibility and rollback.** No existing document shape changes; `RecommendationAudit` gains one optional field and audits written before this branch replay unchanged. `POLICY_VERSION` moves to `2026-08-external-plan-provenance-v1` and the outgoing value joins `HISTORICAL_POLICY_VERSIONS`, so decisions persisted under it stay auditable but are explicitly non-replayable — the intended ADR-0010 behaviour. Rollback is setting `planningMode` back; no migration.

**Also worth knowing:** `check-policy-drift.mjs` now guards `externalSession.ts` and `externalSessionProfiles.ts`. They decide what an externally-planned athlete is told to do, so a change there alters a persisted decision exactly as a change to `rules.ts` does, and the path-based gate did not previously know that.

## Domain invariants

- [x] Firestore writes remain user-scoped under `users/{APP_USER_ID}/...`; no `default_user` or top-level recovery snapshots were introduced. All new paths are `users/{userId}/external_plans/{planId}` plus its `revisions/` (create-only) and `placement/` subcollections, verified by emulator tests including cross-user denial.
- [x] Calendar-date logic uses `Europe/Warsaw`; completed `totalSteps` still means the previous calendar day (`D - 1`). All new date arithmetic goes through `addDaysToLocalDateString` / `getDayDiff` / `getLocalDateString`; no `toISOString().split('T')[0]`. Step semantics are untouched.
- [x] No credentials, tokens, service-account files, or raw health payloads are included. The decision audit gains only a plan id, revision, session id and content hash.
- [x] Recommendation decision logic changed: `POLICY_VERSION` was evaluated and bumped, and `docs/architecture/recommendation-engine.md`, `docs/external-plan-schema.md`, `AGENTS.md`, `docs/README.md` and `docs/plans/README.md` were updated.

## Screenshots or recordings

The visual harness gained a `home-externally-planned` scenario, captured at both viewports into `app/artifacts/visual-review/latest/` (gitignored, regenerable with `cd app && npm run visual:refresh`). Reviewing that capture is what surfaced the raw-enum defect noted above, so it is worth regenerating rather than taking on trust.

The desktop capture shows, top to bottom: the verdict banner leading with **Not today** over the imported session and its plain-language exclusion; the rest template beneath it, so an excluded day cannot read as "do this"; the plan's own week with per-day placement, `Missed it` controls only on past days, and the advisory weekly findings closing with "Observations only. Your plan's author decides what you train; none of this changes a session or its verdict."

---
_Generated by [Claude Code](https://claude.ai/code/session_01PmvVAwr1nxXHqPQTK6yLQg)_


Szczepanov (commented):
Re-review of Phase 8 at head `2ed497b43c2e8572f475846b3eb877a3b8dd12f5`. CI Pipeline #341 is green after the review-driven changes below.

The overall direction is good. The selection/adjudication split is clear, the synthetic-template boundary is constrained, the free-text fallback stays advisory, the weekly critique is non-mutating, and the provenance/replay work is much stronger than a typical first implementation of external plans.

I pushed three small integrity improvements while reviewing:

- `0c835a1` — reject header/revision/placement documents whose embedded identity disagrees with the Firestore path. In particular, replay must not ask for `plan A / revision 2` and receive schema-valid bytes that internally claim to be another plan/revision.
- `72a5695` — reject a placement overlay whose revision no longer matches the active immutable revision instead of silently applying old placement decisions to new session bytes.
- `2ed497b` — focused regressions for both boundaries.

Those are green in the full CI suite and do not change training policy.

## Merge blockers

### 1. Externally-planned sessions do **not** currently pass the same resolved feasibility gates as catalog candidates

This is the most important issue I found.

`evaluateTrainingWithIntent` returns through `adjudicatedExternalRecommendation(...)` immediately after intent resolution. The normal path only afterwards calls `resolveAvailability(...)`, applies fixed-activity stimulus/capacity, and filters/ranks against the resolved day's `maxTimeMinutes`, available equipment and environment override.

`adjudicateExternalSession`, meanwhile, calls `evaluateTemplateEligibility(...)` with raw `readiness.subjective.timeAvailable` and the standing `UserContext`.

Concrete consequence: if the athlete reports 150 min available and has a booked 90 min football session today, a 60–75 min imported workout can still return `proceed`, while the normal catalog path sees only 60 min remaining (and any slightly longer imported session clearly diverges). The same bypass exists for `availabilityOverride` and day-specific environment/equipment context.

That violates ADR-0019 D-CANDIDATE's core promise that the imported session is adjudicated on the same feasibility terms as a catalog candidate.

Recommended shape:

1. resolve today's `ResolvedAvailability` before the external early return;
2. pass that resolved object into the external adjudicator;
3. use its `maxTimeMinutes`, `availableEquipment`, and `environmentOverride` rather than recomputing a weaker approximation;
4. add regressions for fixed-activity time consumption and a day-specific environment/equipment override.

Do not duplicate `resolveAvailability` policy inside `externalSession.ts`.

### 2. D-EVENT is implemented differently from the accepted ADR

ADR-0019 explicitly says an `isEvent` session is reconciled onto the existing `FixedActivity` contract, is **never itself recommended**, and that import must surface a missing matching `UserEvent` for link/create rather than silently inventing one.

The implementation instead short-circuits `externalSession.ts` to `advisory`, converts the event to the synthetic external template, and `externallyPlannedMode.test.ts` explicitly asserts that the event template is returned as the day's recommendation. I did not find the required UserEvent reconciliation/linking path.

This needs an explicit decision before merge: either implement accepted D-EVENT, or amend ADR-0019 + Phase 8 plan and review the new semantics. I would not silently let code redefine an Accepted ADR here.

## High-priority contract gaps

### 3. Empty external-plan days are not visibly labelled as fallback

`resolvePlanningContext` correctly sets `externalFallback`, but the normal ranked-return rationale does not consume it. The test only proves that a catalog pick happens; it does not prove the athlete is told why.

Phase 8.6 requires an unplaced day to produce a **labelled fallback**, and the ADR says the fallback must never be silent. Please make this an explicit recommendation/audit signal or at least a stable rationale prefix, with a regression.

### 4. `supersededFrom` is stored but not used when resolving the active revision

`ExternalPlanService.import(..., supersededFrom)` records the effective-from date, but `ActiveExternalPlanService` simply loads the latest header revision for every covered date. Resolving a date before `supersededFrom` therefore uses the new bytes even though D-IMMUT says a new revision supersedes the old one only from the chosen date forward.

Persisted historical recommendations protect some UI paths, but the domain resolver itself still violates its stated temporal contract. Either model revision effective ranges sufficiently to resolve this correctly, or narrow/remove the forward-only claim.

### 5. D-EXTTIER says modality × intensity × **duration**; the implementation currently ignores duration

`deriveExternalSessionProfiles` selects the modality/intensity fallback vector directly. A 30-minute hard ride and a 180-minute hard ride therefore receive the same cost/stimulus profile before other logic.

The ADR and Phase 8 plan both explicitly include duration in this derivation. Please either consume duration through the existing duration-reference machinery (without inventing a second physiology model) or amend the contract. This matters to both daily adjudication and weekly critique.

## Smaller correctness/UX items worth fixing in this PR

- `proposeReplacement`: when a Monday miss is noticed on Thursday and Thursday itself is free, `earliest = today` followed by `date > earliest` skips Thursday and starts at Friday. The invariant is “never in the past”, so today should be eligible when the miss is reported late.
- Re-import diffing only compares a subset of the session contract. Changes to priority, flexibility/ifMissed, equipment, environment, objectives, scaling, structured steps or `isEvent` can currently be behaviorally important without appearing in the review diff. At minimum, fall back to a deep structural “other session details changed” marker.
- The import textarea placeholder uses `external-training-plan/v1`, while the actual schema constant is `adaptive-training-recommender/external-plan@1`.

## Verdict

Strong Phase 8 foundation, and the branch is substantially better after the integrity fixes. I would **not merge yet** because items 1 and 2 are direct contradictions of ADR-0019's accepted core contracts. Items 3–5 should also be resolved before the plan is honestly called `Implemented`; the remaining smaller items can be handled in the same cleanup pass.

After those changes, I would re-run the external-mode acceptance suite with special attention to: fixed-activity collision days, event days, unplaced fallback days, revision cutover dates, and short-vs-long sessions with otherwise identical modality/intensity.

Szczepanov (commented):
Follow-up to my review at `5948b495`: I addressed the blockers/gaps I raised and re-reviewed the resulting behavior through head `f35ec5dbd409d6f8f3330fe845181cb96b8bf4cb`.

## Status

**All findings from my prior review are resolved. I no longer see a blocker from that review.**

CI Pipeline **#372** (`31937811774`) is fully green on this exact head: Python lint/typecheck/tests/audit, Docker build, frontend lint, workout validation, policy-drift guard, frontend tests + coverage, Firestore emulator, simulation bounds, semantic diff, production frontend build, and Node dependency audit all pass.

## What changed

1. **External sessions now use the same resolved day feasibility as catalog candidates.** `adjudicateExternalSession` accepts the existing `ResolvedAvailability`, and the production external path supplies `resolveAvailability(...)`. This closes the fixed-activity time, day-specific equipment and environment-override bypass. Integration tests cover fixed-activity time and an indoor-only day excluding an outdoor imported session.

2. **D-EVENT is now implemented as written rather than treating the event as a normal external prescription.** An imported `isEvent` session is adjudicated for advice, converted in memory to the existing `FixedActivity` contract, and the normal engine ranks any *additional* recommendation around that commitment. The event itself is never persisted as the recommended synthetic template. A missing compatible `UserEvent` on the date is surfaced explicitly with a link/create instruction; the import still does not fabricate calendar state.

   I also closed the consequences of that change rather than stopping at today's UI:
   - the event's reserved time/load participates in today's availability/fatigue;
   - the event's derived stimulus receives **inferred** (`authoredExternal`) objective credit, not exact catalog credit and not zero anonymous-fixed-activity credit;
   - the in-memory event load survives into tomorrow's projected history without double-counting caller-owned fixed activities;
   - replay distinguishes ordinary externally-selected sessions (no ranking) from event-informed hybrid decisions (external event provenance + normally ranked additional recommendation), while still checking revision/hash/session identity and highest-utility selection.

3. **Empty externally-planned days are labelled fallback.** The ranked-path rationale now starts with an explicit `External plan fallback...` explanation when no imported session is placed today.

4. **`supersededFrom` is now load-bearing.** `ActiveExternalPlanService` will not activate a newly imported revision before its effective-from boundary. A valid placement overlay from an older revision is treated as stale metadata and is not applied to newer immutable bytes; the new revision starts unmodified instead of being corrupted or bricked.

5. **External profile derivation now includes authored duration.** Cost/stimulus still reuse the existing modality × intensity fallback tables, but are scaled relative to the median duration of comparable catalog sessions rather than treating 30 min and 120 min sessions identically. The profiles remain clamped to the existing 0..1 contract, and no AI-supplied load value was introduced.

6. **Late missed-session handling can use today.** If a miss is noticed later and today is free, `reschedule_within_week` / `carry_forward` may propose today. A miss reported on the missed day itself still starts strictly after that date, and occupied/past dates remain excluded.

7. **Re-import diff is behaviorally complete enough for this contract.** It now detects priority, flexibility, `ifMissed`, environment/equipment, objective tags, scaling/fallback, event status and full prescription-structure changes in addition to the previous title/week/day/modality/intensity/duration checks. Set-like array reordering is ignored. The pure diff helper was extracted from the React component so Fast Refresh/lint remains clean.

8. **The import placeholder uses the real schema constant** (`EXTERNAL_PLAN_SCHEMA`) instead of the stale example identifier.

## Additional integrity hardening retained from the first pass

- header/revision/placement embedded identity must agree with its Firestore path;
- replay cannot trust valid plan bytes stored under a different path identity;
- a stale placement revision can never be applied to a newer plan revision.

## Verification notes

The follow-up added focused regressions around all of the above. Two CI-only issues surfaced during the work and were fixed rather than waived: the external-mode integration tests no longer fall through to a real Firestore provider against the dummy CI project, and the production TypeScript build caught/closed structural typing mistakes that Vitest transpilation did not.

From the scope of my original review, this is now in a mergeable state. I would still treat an independent review of the overall 60-file Phase 8 change as useful because of its size, but I do not have an outstanding change request from these findings.

Szczepanov:
## PR Review: Phase 8 (Externally-Planned Mode — Import, Adjudicate, Critique, Replay)

### Executive Summary & Scorecard

This PR is a comprehensive, exceptionally well-architected implementation of **Phase 8** ([ADR-0019](file:///docs/adr/0019-externally-planned-mode.md)). It introduces external training plan imports, strict schema validation, structural session eligibility gating, readiness adjudication, duration-calibrated load profiles, weekly critique diagnostics, immutable revision storage, and hash-verified decision replay.

The follow-up commits (`5948b49` → `f35ec5d`) thoroughly addressed previous self-review blockers (availability resolution on external sessions, D-EVENT fixed activity conversion with inferred stimulus credit, fallback labeling, duration-scaled profile math, and forward-only revision boundaries). All checks, unit tests (81 test files, 958 passed), AST architecture guards, and simulation diffs pass cleanly.

| Metric | Rating / Value | Description |
|---|---|---|
| **Overall Score** | **9.6 / 10** | Production-ready, clean domain separation, high test rigor. |
| **Review Confidence** | **95% (High)** | Verified against codebase, AST import graph tests, and simulation engine. |
| **Urgency** | **Low (Non-blocking)** | Ready to merge. |
| **Priority** | **P3 (Normal)** | Polish, UX edge cases, and Firestore rule defense-in-depth. |

---

### Key Findings & Fixes Applied in `f1f80d1`

1. **`ExternalPlanWeek` Manual Day Picker `FixedActivity` Filtering**:
   - **Score Impact:** +0.2 | **Confidence:** High (95%) | **Urgency:** Low | **Priority:** P3
   - *Issue:* When selecting *"Choose another day"* on a missed session proposal, `occupied` dates only counted imported plan sessions, ignoring booked commitments (e.g. football matches or races).
   - *Fix:* Passed `fixedActivities={planWeekFixedActivities}` into `ExternalPlanWeekProps` and unioned non-completed fixed activity dates into `occupied`.

2. **Proactive / Same-Day Rescheduling from Home UI**:
   - **Score Impact:** +0.2 | **Confidence:** High (90%) | **Urgency:** Low | **Priority:** P3
   - *Issue:* The reschedule button was restricted to `date < today`. An athlete who knew early in the morning that they could not train today (or received a `defer`/`skip` verdict) had to wait until the next day to reschedule.
   - *Fix:* Allowed rescheduling on `date <= today`, displaying `"Reschedule"` for `today` and `"Missed it"` for past dates.

3. **Firestore Security Rules Path & Revision Binding**:
   - **Score Impact:** +0.1 | **Confidence:** High (95%) | **Urgency:** Low | **Priority:** P4
   - *Issue:* While client TypeScript enforced document identity, Firestore rules allowed cross-path writes under a user's own collection without asserting that `data.planId == planId`, `string(data.revision) == revisionId`, or `placementId == 'current'`.
   - *Fix:* Added exact path and ID equality assertions in `firestore.rules` and covered them in emulator tests.

4. **Resilient Proposal State on Failed Save**:
   - **Score Impact:** +0.1 | **Confidence:** Medium (85%) | **Urgency:** Low | **Priority:** P4
   - *Issue:* `setProposal(null)` was called before or unconditionally during save, which discarded the athlete's open proposal dialog if network persistence failed.
   - *Fix:* Wrapped confirm/choose actions in try/catch to retain open proposal state and allow immediate retry on failure.

---

### Verdict

**LGTM — Approved for merge.** All Phase 8 requirements and integrity constraints are fully satisfied.

PR #53, opened 2026-08-16
docs: shadow-mode analysis and Phase 9.0 plan

Szczepanov:
Answers a direct question -- should the manual daily AI loop be replaced by the app now that Phase 8 has landed -- and plans the work that would answer it with evidence instead of taste.

The analysis argues: not yet, and not because the app is too weak. Phase 8 narrowed the engine's job to adjudication, which is where deterministic rules beat a language model, and the cross-day state the app holds is something a fresh chat structurally cannot have. The reservation is narrower than the old one: the athlete's own subjective scores enter the decision path unnormalised, and the corpus that would measure that has zero subjective variance by construction (scenarios.ts builds nearly every case from a constant), so a calibration run today would report "no effect" for a structural reason.

Phase 9.0 is the plan that fixes the evidence gap rather than the engine: run both loops for one block and record where they disagree. It yields the disagreement log, and -- as a by-product of simply using the app -- the real subjective corpus Phase 9's own 9.5 currently has to invent.

Two design commitments are enforced rather than intended: the journal cannot be reached from any decision module (an import-graph guard, the same one-way boundary D-CRITIQUE draws for the weekly critique), because a journal the engine reads is evidence contaminated by its own effect; and anchoring is recorded per-day and reported split rather than assumed away.

Phase 9's 9.5 is re-scoped accordingly: after a shadow block its profiles are parameterised from observed check-ins rather than chosen. Not a hard blocker -- invented variance still beats a constant -- but stated as the better ordering.

No code touched; POLICY_VERSION unchanged. 963 app tests pass on this base.

## Validation

- [ ] `uv run pytest` (backend changes)
- [ ] `uv run ruff check .` and `uv run mypy src/garmin_sync` (backend changes)
- [ ] `cd app && npm run check` (frontend changes)
- [ ] `cd app && npm run test:rules` (Firestore rules changes)
- [ ] `cd app && npm run simulate:scenarios` (recommendation-engine changes)
- [ ] `cd app && npm run visual:refresh` (material UI changes)

## Domain invariants

- [ ] Firestore writes remain user-scoped under `users/{APP_USER_ID}/...`; no `default_user` or top-level recovery snapshots were introduced.
- [ ] Calendar-date logic uses `Europe/Warsaw`; completed `totalSteps` still means the previous calendar day (`D - 1`).
- [ ] No credentials, tokens, service-account files, or raw health payloads are included.
- [ ] Recommendation decision logic changed: `POLICY_VERSION` was evaluated and relevant architecture/ADR documentation was updated.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -85,6 +85,7 @@ Point-in-time assessments of the system as built, including gaps between documen
 * [**2026-08-10 Training-intent & periodization architecture analysis**](./analysis/2026-08-10-training-intent-periodization-architecture.md) — Evidence for Phase 7A/7B mode, preference-ownership, capacity, and stateful-allocation proposals.
 * [**2026-08-11 Phase 6 & 7 compliance review**](./analysis/2026-08-11-phase-6-7-compliance-review.md) — Code, test, simulation, deployment, and plan-status review of the delivered Phase 6 and Phase 7 work.
 * [**2026-08-15 Externally-authored plan feasibility**](./analysis/2026-08-15-externally-authored-plan-feasibility.md) — Structural assessment of importing an externally-authored training plan and narrowing the engine to per-session adjudication.
+* [**2026-08-16 Manual AI loop vs app adjudication**](./analysis/2026-08-16-manual-loop-vs-app-adjudication.md) — Whether the athlete's manual daily AI loop should be replaced by the app now that Phase 8 has landed. Verdict: shadow both for one block first; the subjective input is the one unmeasured path.
 
 ---
 
@@ -101,6 +102,7 @@ How agreed changes get made. Mutable, status-tracked, and expected to go stale �
 * [**Phase 7A: Weekly allocation & safe role reservations**](./plans/phase-7-weekly-allocation-and-role-reservations.md) — The PR #17 follow-up: reserve eligible required roles, protect them from support work, and report safety-forced misses.
 * [**Phase 7B: Training intent, capacity & planning modes**](./plans/phase-7-training-intent-and-planning-modes.md) — Evergreen evidence-to-dose-to-capacity planning proposal.
 * [**Phase 8: Externally-planned mode**](./plans/phase-8-externally-planned-mode.md) — *Implemented.* Import an externally-authored plan, adjudicate its sessions against daily readiness, and repoint the weekly machinery to critique.
+* [**Phase 9.0: Shadow mode and the decision journal**](./plans/phase-9-0-shadow-mode-and-decision-journal.md) — *Draft.* Run the app and the athlete's AI side by side for one block and record where they disagree; produces the real subjective corpus Phase 9 needs.
 * [**Phase 9: Subjective baselines in readiness mode**](./plans/phase-9-subjective-baselines.md) — *Draft.* Build the tighten-only subjective drift term behind a default-off selector, give the corpus real subjective variance, then decide from evidence whether it ships.
 
 ---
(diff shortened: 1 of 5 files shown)


Szczepanov:
## PR Review: Shadow-Mode Analysis & Phase 9.0 Plan

### Executive Summary & Scorecard

This PR introduces an exceptionally well-reasoned architectural analysis ([`docs/analysis/2026-08-16-manual-loop-vs-app-adjudication.md`](file:///docs/analysis/2026-08-16-manual-loop-vs-app-adjudication.md)) and an evidence-first roadmap ([`docs/plans/phase-9-0-shadow-mode-and-decision-journal.md`](file:///docs/plans/phase-9-0-shadow-mode-and-decision-journal.md)) addressing whether the manual daily AI prompt can be replaced by the app following Phase 8.

The thesis is grounded and pragmatically sequenced: **shadow before switching**. While the engine's stateful cross-day memory (fatigue decay, objective ledgers, hard feasibility gates) outperforms conversational AI memory, subjective check-in inputs remain un-normalised on the decision path (the subject of Draft Phase 9 / ADR-0020), and existing scenario fixtures have zero subjective variance. Phase 9.0 establishes a 4–6 week empirical shadow block to produce a real-world disagreement log and an observed subjective check-in dataset to parameterise Phase 9.5.

| Metric | Rating / Value | Description |
|---|---|---|
| **Overall Score** | **9.8 / 10** | Outstanding analytical clarity, sound empirical methodology, clear system boundaries. |
| **Review Confidence** | **95% (High)** | Verified against architecture documents, ADR-0019/ADR-0020, and test invariants. |
| **Urgency** | **Low (Non-blocking)** | Ready to merge. |
| **Priority** | **P3 (Normal)** | Minor document clarifications on entry lifecycle and adherence alignment. |

---

### Key Strengths

1. **Airtight Separation Invariant (9.0.6)**: Requiring an AST runtime-import-graph test guarantees that no module in the decision path can ever import `decisionJournalService.ts` or `shadowLog.ts`, preventing the journal from contaminating the decisions it measures.
2. **Honest Handling of Anchoring Bias (9.0.3 & 9.0.7)**: `sawEngineVerdictFirst` is captured from UI interaction telemetry rather than self-reported checkboxes, and reporting is explicitly partitioned by this flag.
3. **Formal Conservatism Ladder (9.0.4)**: Classifying agreement in pure code across the five shared verdict literals prevents subjective bias during analysis.
4. **Pragmatic Sequencing for Phase 9.5**: Re-scoping Phase 9.5 to sample observed standard deviations and baselines from the shadow block solves the synthetic zero-variance fixture limitation without blocking Phase 9.0.

---

### Follow-up Refinements Applied in `2e61076`

1. **Operational Schedule (9.0.1)**: Explicitly specified `05:00 Europe/Warsaw` cron schedule (`0 5 * * *`) for Cloud Scheduler so daily sync and snapshot generation are completed before morning check-ins.
2. **Journal Mutation Lifecycle & Immutability (9.0.2)**: Documented morning creation vs evening update flow, enforcing immutable `createdAt` and `sawEngineVerdictFirst` to protect anchoring telemetry.
3. **Adherence Alignment (9.0.3 / 9.0.5)**: Clarified that `actualVerdict` defaults naturally from the daily `AdherencePrompt.tsx` response (`followed === true`) to eliminate redundant double-entry fatigue while retaining optional manual overrides.

---

### Verdict

**LGTM — Approved for merge.** The analysis and plan provide clear, disciplined direction for validating daily adjudication with empirical evidence.

PR #55, opened 2026-08-16
feat(phase-9-0): decision journal storage and agreement classification

Szczepanov:
## Summary

Completes and hardens the **Phase 9.0 code slice (9.0.2–9.0.6)** for running the app beside the athlete's existing AI planning loop and collecting evidence without letting that evidence influence recommendations.

### Decision journal and evidence integrity

- Adds `ShadowVerdict` / `DecisionJournalEntry` and owner-scoped `users/{uid}/decision_journal/{date}` persistence.
- Morning evidence is **create-only and non-deletable**: `externalVerdict`, `externalNote`, `sawEngineVerdictFirst`, and `createdAt` cannot be rewritten after the engine may have been revealed. Only the evening `actualVerdict` / `updatedAt` may change.
- INVALID/UNAVAILABLE stored journal evidence blocks a new morning write instead of being treated as a missing day.
- The persistence parser enforces the closed key set, schema version, owner/date identity, verdict enums, timestamps, and the 2000-character note bound.
- Range export surfaces invalid journal records as data-quality gaps instead of silently reducing the sample.

### Anchoring measurement

- Home can keep Today's Recommendation hidden while the athlete records their own planner verdict.
- `sawEngineVerdictFirst` comes from observed interaction, not a self-report checkbox.
- The observed reveal is remembered locally per user/date across same-day navigation/reload so refreshing the page cannot manufacture a later "blind" row when browser storage is available.
- Once a morning entry exists it is displayed read-only; only the actual outcome remains editable.

### Exact engine verdict

- New recommendation writes persist the exact five-value `engineVerdict` (`proceed | scale | defer | skip | advisory`) as evidence metadata.
- Imported-plan `skip` and target-event `advisory` decisions therefore survive persistence/history parsing/export instead of being reconstructed incorrectly from `train | modify | recover`.
- Legacy rows without the field remain readable through the documented mode fallback. First exact-verdict backfill does **not** fabricate a historical decision revision; once the exact field exists, changing it follows the normal archive/revision contract.
- The live Decision Journal uses the same exact resolver as the export.
- Adherence can fill a missing actual verdict from the exact persisted recommendation when `followed === true`, except for `advisory`, which is deliberately a non-instruction and cannot identify what action actually happened.

### Agreement and export

- `engine/shadowAgreement.ts` classifies all 25 ordered verdict pairs on `proceed > scale > defer ≈ skip`; `advisory` is incomparable in both directions.
- `engine/shadowLog.ts` + `services/shadowLogService.ts` join recommendation, journal, subjective check-in, recovery deltas, adherence, policy version and external-plan content hash into a per-day evidence row / CSV.
- Partial days stay visible as partial rows; raw wearable payloads, identifiers and check-in free text are excluded.

### Engine isolation

- Extends the runtime import-graph guard so `rules.ts`, `optimizer.ts`, `planner.ts` and `trainingIntent.ts` cannot reach the decision journal or shadow export.
- `POLICY_VERSION` remains unchanged because this PR changes evidence capture/persistence/UI, not recommendation selection or safety policy.

### Test/CI hardening

- Adds service/parser/export/agreement/recommendation persistence regressions for immutable morning evidence, exact `skip`/`advisory`, legacy verdict backfill, invalid-record surfacing and revision semantics.
- Firestore emulator coverage proves owner scoping, immutable morning fields, legitimate evening updates, exact-engine-verdict validation and journal deletion denial.
- Removes an unrelated nondeterministic `contextBriefService` test leak to real Firestore by mocking the remaining goal-service dependency.

## Remaining Phase 9.0 work

- **9.0.1** unattended Garmin ingestion/backfill/audit is operational and still outstanding.
- **9.0.7** run the 4–6 week shadow block after 9.0.1 is proven.
- **9.0.8** write the dated readout and decide whether to switch from the manual daily loop.

The journal cannot be backfilled: a genuinely blind historical observation cannot be reconstructed.

## Validation

Final head: `740c3267add17a23bc685d9389979b39a28577fe`

**CI Pipeline #412 / run 31949959191: PASS.**

- Frontend lint: pass
- Workout validation: pass
- Engine policy-drift guard: pass (`POLICY_VERSION` unchanged as intended)
- Frontend unit suite + coverage: pass
- Firestore emulator rules: pass
- Simulation scenarios + aggregate bounds: pass
- Simulation semantic diff: pass
- Production frontend build/typecheck: pass
- Node dependency audit: pass
- Python lint/typecheck/pytest/dependency audit: pass
- Docker image build: pass

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01E6MPwCuFPdGEWuxtjsMYAv


Szczepanov (commented):
Follow-up review after the Phase 9.0 hardening pass.

I reviewed the code path end-to-end against the stated purpose of this phase: collect comparison evidence without letting that evidence contaminate the engine. I also applied the fixes directly on this branch. I no longer see a code-level merge blocker on the current head.

### Issues found and resolved

1. **Blind journal evidence could be rewritten after seeing the engine.** A row could retain `sawEngineVerdictFirst=false` while the athlete edited the morning verdict/note after reveal. Morning evidence is now create-only: `externalVerdict`, `externalNote`, `sawEngineVerdictFirst`, and `createdAt` are immutable. Only the evening actual outcome can change.

2. **Journal rows could be selectively deleted.** That would make disagreements erasable after the fact and bias the sample. Decision-journal deletion is now denied in Firestore and removed from the service API. Emulator tests cover the denial.

3. **A same-day reload could forget an earlier engine reveal.** The reveal marker previously lived only in React state, so reload/navigation could manufacture a later “blind” entry. The observed reveal is now remembered locally per user/date and re-read at submission time. Current-page behavior remains correct even if browser storage is unavailable; the plan/readout should still treat anchoring as measured evidence rather than assume independence.

4. **The export lost the exact five-value engine verdict.** Reconstructing from `train|modify|recover` mislabeled imported-session `skip` and target-event `advisory` decisions (for example, `advisory + train` became `proceed`). New recommendation writes now persist the exact `engineVerdict`; history parsing, the live journal, and shadow export use it. Legacy rows without the field retain the documented mode fallback only.

5. **Exact-verdict revision semantics could disagree between the service and Firestore.** First backfill onto a legacy row now completes missing metadata without fabricating a historical revision; once an exact verdict exists, changing it is a real decision change and follows the existing archive/revision contract. Tests cover both paths.

6. **Malformed historical journal rows were silently omitted.** That could make sample coverage look better than it was. The journal parser is stricter (closed keys/schema/owner/date/enums/note bound), range reads count invalid records, and the export surfaces them in `unavailableSources` rather than treating them as genuine missing days.

7. **Adherence could synthesize an outcome from `advisory`.** `advisory` is a non-instruction, so “followed” cannot determine what action actually occurred. Auto-fill now uses the exact persisted verdict but deliberately leaves advisory actual outcomes for the athlete to record.

8. **CI had an unrelated nondeterministic Firestore leak.** `contextBriefService.test.ts` still reached the real goal service; that dependency is now mocked, making the suite deterministic.

### Validation

Final head: `740c3267add17a23bc685d9389979b39a28577fe`

**CI Pipeline #412 / run 31949959191: PASS**

- frontend lint, typecheck/unit suite/coverage and production build: pass
- workout validation + policy-drift guard: pass
- Firestore emulator rules: pass
- simulation scenarios + semantic diff: pass
- Node dependency audit: pass
- Python lint/typecheck/pytest/dependency audit: pass
- Docker image build: pass

The Phase 9.0 plan and PR description were also updated to match the implemented contract; they no longer claim `skip/advisory` are only approximated or that 9.0.3/9.0.5 are future work.

### Remaining work is intentionally outside this code review

- **9.0.1:** prove unattended Garmin ingestion/backfill/audit operationally.
- **9.0.7:** run the 4–6 week shadow block after 9.0.1 is proven.
- **9.0.8:** analyze the resulting block and make the workflow decision.

Those are prerequisites for trusting the eventual evidence, but I do not consider them merge blockers for the 9.0.2–9.0.6 code slice itself.

**Verdict:** code-level issues I found are addressed; current head is mergeable from this review’s perspective. I did not merge the PR.

PR #58, opened 2026-08-16
Phase 9.3: subjective drift term behind a default-off selector

Szczepanov:
## Summary

Implements **Phase 9.3** from ADR-0020: a subjective-drift candidate inside the real readiness evaluator, behind a default-off selector.

- Adds `SubjectiveDriftPolicy = 'off' | 'drift'` and `subjectiveDriftStrain` in `rules.ts`.
- Under `'drift'`, the current reference candidate signs adverse recent-vs-long movement positive, normalizes by the baseline variability estimate, floors favourable movement at zero, caps per metric, and aggregates non-negative weighted contributions.
- `evaluateReadinessAndSafetyEnvelope` defaults the selector to `'off'`; existing production call sites are unchanged, so the new mechanism is dormant and cannot affect persisted recommendations.
- Absolute subjective thresholds remain hard floors: drift can only escalate `train -> modify -> recover`.
- The estimator/weights remain experimental policy inputs for Phase 9.6; this PR does **not** claim they are physiologically calibrated.

## Review adjustments

A manual review found and fixed two provenance issues:

1. **No `POLICY_VERSION` bump while dormant.** The original PR bumped the policy version simply because `rules.ts` changed. That contradicted accepted ADR-0020, which explicitly says a default-off implementation must retain the existing policy identity until the live deciding path can change. `POLICY_VERSION` therefore remains `2026-08-external-plan-provenance-v1`.
2. **The policy-drift guard now checks the actual version value.** Previously, merely touching `policy.ts` satisfied the guard even if `POLICY_VERSION` did not change. The script now compares the base and head values, supports the narrowly-defined ADR-0020 dormant-selector exception, and includes `subjectiveBaseline.ts` in the decision-affecting set so post-ship estimator changes cannot silently retain an old policy identity.

## Validation / safety properties

- `'off'` remains the production default.
- No production source file outside the dormant subjective-drift implementation is changed by this PR.
- Unit/property coverage exercises adverse/favourable movement, per-metric polarity, caps, zero/negative weights, absolute-floor preservation, and the no-de-escalation invariant.
- Simulation semantic diff is expected to show no changed pre-existing production scenario while the selector is off.
- Final CI on the reviewed head is the source of truth before merge.

## Deliberately deferred

- **9.4:** composition-boundary validated subjective-history read.
- **9.6:** estimator/parameter sensitivity comparison.
- **9.7:** drift-inclusive `DecisionScoreTelemetry`, compact audit provenance, replay/rationale integration. Until 9.7 lands, `totalDecisionScore` remains objective-only even when tests explicitly invoke `'drift'`; 9.6 must not treat that field as a drift-inclusive total or duplicate the formula silently.
- **9.8:** production activation decision, still requiring both simulation/regression evidence and prospective Phase 9.0 evidence.

No user-visible behavior changes in this PR.


Szczepanov:
Reviewed and adjusted.

Key findings fixed:

- Restored ADR-0020 policy-version semantics: the Phase 9.3 implementation is still default-off and cannot alter a persisted recommendation, so `POLICY_VERSION` correctly remains `2026-08-external-plan-provenance-v1` rather than creating a synthetic dormant policy identity.
- Hardened `check-policy-drift.mjs`: it now compares the actual base/head `POLICY_VERSION` values instead of treating any edit to `policy.ts` as a version bump.
- Added the narrow Accepted-ADR-0020 dormant-selector exception, while keeping normal live decision changes subject to a real version bump.
- Added `subjectiveBaseline.ts` to the decision-affecting set so, after any future live cutover, estimator changes cannot silently keep an old policy identity.
- Reconciled the Phase 9 plan with the already-reviewed 9.1/9.2 implementation: six-score subjective completeness is the baseline coverage authority, the old `models.ts` ↔ `subjectiveBaseline.ts` type cycle is gone, and the 9.3 no-bump rule is documented correctly.
- Documented the current telemetry boundary explicitly: when tests/measurement invoke `'drift'`, `DecisionScoreTelemetry.totalDecisionScore` is still objective-only until 9.7. A 9.6 harness must not interpret it as drift-inclusive or duplicate the drift formula in a second implementation.

I re-reviewed the Phase 9.3 drift mechanics. For a dormant experimental reference candidate, the shape is sound: favourable relative movement cannot subtract, absolute subjective thresholds remain authoritative floors, and production callers still default to `'off'`. I deliberately did not recalibrate metric weights/caps here; ADR-0020 assigns estimator/parameter sensitivity to 9.6 and requires prospective Phase 9.0 evidence before any production activation.

Final head: `6d69fa1a80443e54a58fc4856f16761af6541b8a`

CI Pipeline #439 is fully green on that head: frontend tests/coverage, strengthened policy-drift guard, Firestore emulator, simulation bounds + semantic diff, production build, Node audit, Python suite/audit, and Docker all passed.

I do not see a remaining merge blocker for this dormant Phase 9.3 PR. I did not merge it.

PR #67, opened 2026-08-17
feat(ui): add expandable workout details in external plan week

Szczepanov:
### Summary

- Adds expandable 'View workout' / 'Hide workout' button to each session row in \ExternalPlanWeek.tsx\.
- Displays structured workout prescription steps (sets, intervals, durations, targets, notes) and author fallback notes.
- Extracts shared step timing formatting logic to \externalPrescriptionUtils.ts\.
- Adds unit tests and updates visual capture test selectors.

### DIFF
--- a/app/src/components/ExternalPlanWeek.css
+++ b/app/src/components/ExternalPlanWeek.css
@@ -83,6 +83,12 @@
   flex: 1;
 }
 
+.external-week-session-item {
+  display: flex;
+  flex-direction: column;
+  gap: 0.25rem;
+}
+
 .external-week-title {
   font-size: 0.88rem;
   font-weight: 500;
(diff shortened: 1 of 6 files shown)


Szczepanov (commented):
Reviewed and adjusted.

The original new test only rendered the closed state, so it could not catch a broken expand handler. Fixed in 54377b4: the visual flow now asserts that clicking **View workout** changes `aria-expanded` and exposes the labelled workout-details panel before capturing it.

Validated with `cd app && npm run check`:
- TypeScript and ESLint clean
- 95 Vitest suites passed (1,193 tests)
- Workout catalog validation passed

Visual Playwright capture remains blocked in this checkout before any scenario renders: both the externally planned case and unchanged `home-normal-load` stay on “Loading dashboard…” until the existing 15-second readiness timeout. That prevents end-to-end screenshot confirmation here, but it is not caused by this PR’s expanded-workout change.

No remaining blocking findings in the PR diff.

PR #71, opened 2026-08-17
feat(firebase): enable persistent offline cache for gym-floor writes (S1.1)

Szczepanov:
## Summary

- `getDb()` had no local cache configured — every Firestore write required live connectivity. Fine for existing desk-chair writes (check-ins, preferences); wrong for strength set logging, which happens on a gym floor with no signal (ADR-0021 / strength-session-logging plan, S1.1).
- Configures `persistentLocalCache` with the multi-tab manager, preserving `ignoreUndefinedProperties`. Writes now land in IndexedDB immediately and flush on reconnect.
- Added beyond the literal change: `firebase.test.ts` (new) pins the durability contract — every existing service test mocks `../firebase` wholesale, so nothing previously exercised `getDb()` at all; the persistent cache could have been silently removed without a single test failing. The pre-existing bare `catch {}` fallback now warns before dropping to `getFirestore()` (no offline durability) instead of degrading silently.
- No user-visible change beyond durability; this is the first item in a 12-item stacked strength-logging plan (`docs/plans/strength-session-logging.md`).

## Validation

- [x] `cd app && npm run check` — pass: 1212 tests, includes new `firebase.test.ts` (6 tests).
- Manual check: cold-cache caveat verified by inspection — every service test mocks `../firebase`; emulator suite (`firestoreRules.emulator.test.ts`) initializes via `@firebase/rules-unit-testing`, never touching `src/firebase.ts`. No test asserts on cold-cache behaviour, so none needed updating.
- [ ] Not verified: "a write issued while offline is visibly queued and flushes on reconnect" is real-browser/IndexedDB behaviour a Node test can't exercise. Recorded as an open manual-verification item in the plan (DevTools → Offline → write → reload → reconnect) rather than claimed.

## Risk and reviewer guidance

Low risk, one call-site change (`initializeFirestore`'s options object). The property to check: enabling persistence changes read timing for every collection in the app, not just a new one — verified no existing test depends on cold-cache timing (see Validation). Rollback is trivial: revert one options object.

## Domain invariants

- [x] Firestore writes remain user-scoped — unchanged, this only touches client-side cache configuration.
- [x] Calendar-date logic uses Europe/Warsaw — not touched by this PR.
- [x] Recommendation decision logic changed: no. Pure infrastructure change, no `POLICY_VERSION` impact.

## Screenshots or recordings

### DIFF
--- a/app/src/firebase.test.ts
+++ b/app/src/firebase.test.ts
@@ -1,0 +1,96 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/** S1.1 pins the offline-durability contract in `getDb()`. Every service test mocks
+ *  `../firebase` wholesale, so without this file nothing exercises the real
+ *  initialization path and the persistent cache could be dropped without a single test
+ *  turning red -- on a feature whose entire purpose is surviving a gym with no signal. */
+
+const initializeFirestore = vi.fn<(app: unknown, settings: Record<string, unknown>) => { tag: string }>(
+    () => ({ tag: 'persistent-db' }),
+);
+const getFirestore = vi.fn(() => ({ tag: 'fallback-db' }));
+const persistentLocalCache = vi.fn((settings: unknown) => ({ kind: 'persistent', settings }));
+const persistentMultipleTabManager = vi.fn(() => ({ kind: 'multi-tab' }));
+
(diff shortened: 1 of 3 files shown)


Szczepanov:
S1.1 review complete. The persistent local cache initialization and regression tests are sound, so I did not change this branch. The remaining acceptance gate is the planned real-browser offline/refresh test; later stack fixes now use Firestore metadata rather than treating local-cache completion as server synchronization.

PR #72, opened 2026-08-17
feat(strength): strength_sessions schema and read-side parser (S1.2)

Szczepanov:
## Summary

- Implements D-GAUGE and D-SETLOG (ADR-0021). Adds `IntensityGauge`, `LoggedSet`, `LoggedExercise`, `StrengthSessionState` and `StrengthSession` to `engine/models.ts`, and `parseStrengthSession` to a new `persistence/parsers/strengthSession.ts` mirroring `parseNormalizedGarminActivity`'s house style.
- Two fields added beyond the plan's illustrative schema block: `userId` (every other client-writable collection duplicates it for `hasOwnedUserId`/`keepsOwnership`) and `updatedAt` (D-SETLOG requires per-set persistence, so the document needs a monotonic revision field — the role `syncedAt` plays for `NormalizedGarminActivity`).
- Degradation policy is two-tier: a malformed top-level field (date, state, schemaVersion, startedAt, updatedAt, userId mismatch) invalidates the whole document; a malformed *set or exercise* is dropped from its parent array and the rest of the session still parses `AVAILABLE` — the raw log should survive one bad entry from a flaky offline write, not vanish along with every other set.
- `parseWeightKg` is deliberately asymmetric: explicit `null` is bodyweight and accepted; a non-null, non-numeric weight drops the set rather than being coerced to `null`, since that coercion would misreport a real loaded set. No conversion between `rir`/`rpe_rts` happens at parse time, per D-GAUGE.

## Validation

- [x] `cd app && npm run check` — pass: 1232 tests (20 new in `strengthSession.test.ts`).

Results:
- Explicit tests: degradation policy (dropped-set, dropped-exercise cases), all four gauge scales round-trip untouched, `userId` mismatch rejected, `schemaVersion !== 1` rejected.

## Risk and reviewer guidance

Please double-check the two additions beyond the plan's schema (`userId`, `updatedAt`) against S1.3 (Firestore rules, next PR in this stack) — they were added specifically because S1.3's rules design requires them. No engine/recommendation code path reads this collection yet, so this PR has zero effect on any decision.

## Domain invariants

- [x] Firestore writes remain user-scoped — `strengthSession.ts` requires `userId` in the parsed document and checks it against the caller-supplied owner (`user-id-mismatch` otherwise), matching `decisionInputs.ts`'s `parseDecisionJournalEntry` precedent.
- [x] Calendar-date logic uses Europe/Warsaw — `date` validated via the existing `isValidDate`; date *computation* lands in S1.4 (next-next PR).
- [x] Recommendation decision logic changed: no. New types and a parser only; nothing reads this collection into the engine yet.

## Screenshots or recordings

### DIFF
--- a/app/src/engine/models.ts
+++ b/app/src/engine/models.ts
@@ -1255,6 +1255,67 @@ export interface DailyRecommendation {
     };
 }
 
+/** Discriminated intensity gauge for one logged strength set (ADR-0021 D-GAUGE). `rir` and
+ *  `rpe_rts` measure the same failure-proximity construct in different vocabulary and are
+ *  mutually convertible -- but only at READ time; the persisted value is always exactly
+ *  what the athlete entered. `velocity_loss` and `technical` measure output quality, not
+ *  proximity to failure, and are never convertible to or from the other two. Consumers
+ *  that need "how close to failure" (S2.1's 1RM estimator) must branch on `scale`, not
+ *  assume every gauge answers that question. */
+export type IntensityGauge =
+    | { scale: 'rir'; value: number }
+    | { scale: 'rpe_rts'; value: number }
+    | { scale: 'velocity_loss'; percent: number }
+    | { scale: 'technical'; met: boolean; note?: string };
+
+/** One logged set (ADR-0021 D-SETLOG: this is the raw source of truth, never derived
+ *  data written back). `completedAt` is what makes rest duration free -- consecutive
+ *  sets' timestamps are the rest interval; no separate rest field is stored. */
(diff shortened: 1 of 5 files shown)


Szczepanov:
S1.2 review complete and improved in commit 50b62ad. The parser now enforces document-ID/session-ID equality, Warsaw date attribution, lifecycle/timestamp invariants, bounded exercise/set payloads, integer numeric fields, valid gauges, and deterministic duplicate-set handling. Added a shared validation authority and expanded parser regressions.

PR #73, opened 2026-08-17
feat(firestore): strength_sessions rules and validation (S1.3)

Szczepanov:
## Summary

- Adds `hasValidStrengthSession(userId, sessionId)` and the `users/{userId}/strength_sessions/{sessionId}` match block, following `hasValidDecisionJournalEntry`'s pattern. Ownership via `hasOwnedUserId`, `schemaVersion` pinned to 1, `state` restricted to the three literals, `date` validated via the existing `isValidActivityDate`/`isRealCalendarDate` helpers. Update enforces `keepsOwnership` plus immutable `date`/`startedAt` (ADR-0021: date is fixed at session start, never recomputed). Delete is allowed — this is the athlete's own workout log, not an audited record like `decision_journal`.
- **Discovered constraint the plan didn't anticipate**: Firestore's rules language has no per-element predicate over a variable-length `list` — only `is list` and `.size()`. It cannot express "every set's weightKg is in 0..1000" for an array of unknown length. Every existing array-of-maps field in this file (`external_plans`' `sessions`, `plan_blocks`' `assignments`) is already bounded the same way — precedent, not a shortcut. `exercises` is capped at 30; per-set/per-exercise field correctness is the read side's job (S1.2's `parseStrengthSession`, previous PR, already drops a malformed set/exercise rather than trusting it).
- Also added beyond the plan: `sessionId` required and checked against the path segment (`isStrengthSessionDocument`), matching `isDateDocument`'s precedent for every other date-keyed collection.

## Validation

- [x] `uv run pytest` (backend changes) — not applicable, no backend code changed.
- [x] `cd app && npm run test:rules` — **57 emulator tests pass against the real Firestore emulator** (not just typechecked), including 7 new strength-session cases: owner create/read/update/delete, cross-user denied, schemaVersion/state/unknown-key rejection, 30-exercise cap, date/startedAt immutability on update.
- [x] `cd app && npm run check` — pass: 1232 tests, 103 files.

## Risk and reviewer guidance

The load-bearing finding here is #3 in the Summary — please sanity-check the claim that Firestore rules cannot do per-element list validation against your own understanding of the rules language; the whole scope of "what's enforced in rules vs what's enforced by the read-side parser" rests on it. `npm run test:rules` was run against the real local emulator, not just typechecked — worth confirming that ran clean in CI too.

## Domain invariants

- [x] Firestore writes remain user-scoped under `users/{userId}/...` — `hasOwnedUserId` + `keepsOwnership`, tested cross-user in the emulator suite.
- [x] Calendar-date logic uses Europe/Warsaw — `date` validated via `isValidActivityDate`/`isRealCalendarDate` (real calendar-date check, leap years included); immutability on update tested.
- [x] Recommendation decision logic changed: no. Rules only; `POLICY_VERSION` untouched.

## Screenshots or recordings

### DIFF
--- a/app/firebase.json
+++ b/app/firebase.json
@@ -14,7 +14,8 @@
     ]
   },
   "firestore": {
-    "rules": "firestore.rules"
+    "rules": "firestore.rules",
+    "indexes": "firestore.indexes.json"
   },
   "emulators": {
     "firestore": {
--- a/app/firestore.indexes.json
+++ b/app/firestore.indexes.json
@@ -1,0 +1,21 @@
+{
+  "indexes": [
+    {
+      "collectionGroup": "strength_sessions",
+      "queryScope": "COLLECTION",
+      "fields": [
+        { "fieldPath": "state", "order": "ASCENDING" },
+        { "fieldPath": "startedAt", "order": "ASCENDING" }
+      ]
+    },
+    {
+      "collectionGroup": "strength_sessions",
+      "queryScope": "COLLECTION",
+      "fields": [
+        { "fieldPath": "state", "order": "ASCENDING" },
+        { "fieldPath": "startedAt", "order": "DESCENDING" }
+      ]
+    }
+  ],
+  "fieldOverrides": []
+}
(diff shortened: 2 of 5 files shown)


Szczepanov:
S1.3 review complete and improved in commit ecf48f1. Creates must now begin in_progress, completedAt must agree with state, only in-progress sessions may update, and terminal sessions are immutable. Added emulator coverage plus the compound indexes required by active-session queries.

PR #74, opened 2026-08-17
feat(strength): session state machine and date attribution (S1.4)

Szczepanov:
## Summary

- Implements the ADR-0021 lifecycle rules: `date` is fixed at session start (Warsaw-local) and never recomputed; `in_progress` moves freely, `completed` and `abandoned` are terminal.
- Split pure logic from I/O, per D-SUBJPURE's precedent (ADR-0020) of keeping engine-layer rules synchronous and unit-testable without mocking Firestore: `engine/strengthSessionLifecycle.ts` (`computeSessionDate`, `buildNewStrengthSession`, `canTransitionStrengthSessionState`, `isStaleInProgressSession` — 13 tests, no I/O) and `services/strengthSessionService.ts` (`startSession`, `transitionState`, `reconcileStaleSessions` — 12 tests, mocked Firestore).
- Transition rule: `completed`/`abandoned` are terminal against each other and against reopening to `in_progress` — a closed session may already have fed overload history or (once Step C ships) engine credit, so "reopen" is a modelling error (log a fresh session), not a supported transition. Re-saving a terminal state at itself is allowed (idempotent) — matters because `reconcileStaleSessions` processing a session twice must not error.
- `STALE_IN_PROGRESS_HOURS = 6` is this implementation's own choice (the plan specified the requirement, not a number), documented with rationale at the constant.
- Doc ids are minted client-side via an unwritten `doc()` ref (works fully offline, no server round-trip) and embedded as `sessionId` in the single create write the rules require to already agree with the path.
- Note: this branch includes one merge commit (`Merge branch 'main' into feat/s1-4-session-lifecycle`) — that was a manual merge on my part during development, pulling in unrelated upstream work (Zwift/workout-export) that had landed on `main` in the interim. Not part of this PR's actual content.

## Validation

- [x] `cd app && npm run check` — pass: 1262 tests, 105 files.
- Explicit tests: midnight-crossing date attribution in both DST directions (CEST August, CET January) via `Intl.DateTimeFormat`, not a hardcoded offset; all illegal transitions rejected; an abandoned session keeps every set it already logged (never deletes).

## Risk and reviewer guidance

The transition-legality rule (`completed`/`abandoned` terminal, no reopening) is a real design decision, not just plumbing — worth confirming you agree reopening a closed session should be disallowed rather than merely discouraged. `STALE_IN_PROGRESS_HOURS = 6` is an arbitrary but documented choice; change the constant if you'd prefer a different threshold.

## Domain invariants

- [x] Firestore writes remain user-scoped under `users/{APP_USER_ID}/...` — `startSession`/`transitionState` write to `users/{userId}/strength_sessions/{sessionId}` only.
- [x] Calendar-date logic uses Europe/Warsaw — `computeSessionDate` uses `getLocalDateString`, never `toISOString().split('T')[0]`; tested explicitly across both DST directions.
- [x] Recommendation decision logic changed: no. Session lifecycle only; no engine/recommendation code path reads this collection yet.

## Screenshots or recordings

Not applicable — no UI change (UI lands in S1.5, next PR).


Szczepanov:
S1.4 review complete and improved through commits ec62278 and 7bc20b1. Lifecycle construction now rejects invalid instants and blank identities. Terminal transitions are idempotent, reject regressing timestamps, and merge lifecycle fields only, preventing a concurrent set write from being overwritten.

PR #75, opened 2026-08-17
feat(strength): confirm-or-amend set-entry UI (S1.5)

Szczepanov:
## Summary

- Persist-on-every-set, prefill-from-previous-set, confirm-or-amend against a matched prescription. Split three ways:
  - `workouts/strengthSessionEntry.ts` — pure logic, no I/O or React: `extractPlannedStrengthExercises` (prescription matching), `prefillNextSet`, `buildLoggedSet`, `upsertExercise`, `appendSetToExercise`. 20 tests.
  - `hooks/useStrengthSessionRunner.ts` — thin orchestration over the pure module and `strengthSessionService`. No dedicated test file, matching this repo's existing precedent (`useGarminSyncStatus` has none either) and the absence of `@testing-library/react` in this toolchain.
  - `components/StrengthSessionRunner.tsx` — 6 `renderToStaticMarkup` smoke tests in the `GarminSyncBadge.test.tsx` style (no interaction-testing library exists here, so these assert rendered output across states, not clicks).
- Prescription matching reads `WorkoutPrescription.adjustedBlocks` (the original structured `WorkoutBlock[]`), not `displayBlocks` (rendered display strings, unparseable back into structured targets). Of `IntensityTarget`'s eight variants, only `reps_in_reserve` and `technical_quality` map to a suggested gauge — verified against `catalog/support-strength.ts` as the only two a real strength step emits; everything else degrades to no suggestion rather than a guess.
- Wired into navigation: `Screen` gains `'strength'`, routed in `App.tsx`, reachable from `MobileNav`'s "More" drawer at the same tier as "Import Training Plan" / "Export Context for AI". Deliberately **not** added to the primary bottom-nav bar or a `Home.tsx` dashboard card — placement/prominence left for your review rather than assumed.

## Validation

- [x] `cd app && npm run check` — pass: 1292 tests, 107 files.

Results:
- 20 tests for the pure entry-logic module; 6 smoke tests for the component across loading/empty/planned/logged-sets/completed/error states.
- [ ] Not verified: "kill the app mid-session, reopen, every set restored" and "prescribed session prefills from the recommendation" in a real browser — both depend on IndexedDB/persistence behaviour a Node test can't exercise. `logSet` awaits the Firestore write before updating local state, which is the closest available proxy.

## Risk and reviewer guidance

This is the largest PR in the stack (new hook, new component, new pure-logic module, nav wiring). The one thing worth a deliberate look: nav placement is intentionally minimal (More-drawer only) — flag if you want it promoted to the bottom bar or a Home card; I didn't want to make that UX call unilaterally.

## Domain invariants

- [x] Firestore writes remain user-scoped — writes go through `strengthSessionService`, unchanged from S1.4.
- [x] Calendar-date logic uses Europe/Warsaw — session start date is fixed by S1.4's `computeSessionDate`; this PR doesn't add new date logic.
- [x] Recommendation decision logic changed: no. Logging UI only; still no engine/recommendation code path reads `strength_sessions`.

## Screenshots or recordings

Not captured — no live browser session available in this environment to authenticate against Firebase Auth. Recommend a manual pass before merge: `npm run dev`, log in, navigate to Strength via the More drawer.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -8,6 +8,7 @@ import { TrainingSettings } from './components/TrainingSettings';
 import { Preferences } from './components/Preferences';
 import { DataView } from './components/DataView';
 import { ExternalPlanImport } from './components/ExternalPlanImport';
+import { StrengthSessionRunner } from './components/StrengthSessionRunner';
 import { decisionComposer } from './engine/composer';
 import type { DailyDecisionInput } from './engine/models';
 import type { Screen } from './types/navigation';
@@ -114,6 +115,10 @@ function App() {
         
         {screen === 'preferences' && (
           <Preferences userId={userId!} onNavigate={handleNavigate} />
+        )}
+
+        {screen === 'strength' && (
+          <StrengthSessionRunner userId={userId!} />
         )}
 
         {screen === 'plan' && (
--- a/app/src/components/MobileNav.tsx
+++ b/app/src/components/MobileNav.tsx
@@ -92,7 +92,7 @@ export const MobileNav: React.FC<MobileNavProps> = ({ screen, handleNavigate, lo
 
         <button
           ref={mobileMoreBtnRef}
-          className={`nav-item ${['constraints', 'preferences', 'data', 'plan'].includes(screen) ? 'active' : ''}`}
+          className={`nav-item ${['constraints', 'preferences', 'data', 'plan', 'strength'].includes(screen) ? 'active' : ''}`}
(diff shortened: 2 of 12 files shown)


Szczepanov:
S1.5 review complete and improved in commit dfbcb51. The runner now reports synced/pending/unavailable from Firestore hasPendingWrites metadata, validates nested set payloads before the permissive array-rule boundary, keeps gauge draft state single-sourced, and serializes start/select/log/finish actions to avoid UI races.

PR #76, opened 2026-08-17
feat(strength): timestamp-based rest timer and derived rest (S1.6)

Szczepanov:
## Summary

- `workouts/restTimer.ts`: `elapsedSeconds`, `formatElapsed`, `deriveRestSecondsBetweenSets`, `deriveRestIntervals` — pure, 11 tests. `hooks/useElapsedSeconds.ts`: live "time since X" display. No test file, same rationale as `useStrengthSessionRunner` (no `@testing-library/react`; the value it displays is what's actually tested).
- The `setInterval` in `useElapsedSeconds` is the re-render *trigger*, never the accumulator. Every tick recomputes `elapsedSeconds(sinceIso, now)` from wall-clock time, so a throttled or suspended background timer (phone in a pocket between sets) cannot make the display drift low — whichever tick does fire shows the true elapsed time, not a count of ticks that happened to fire.
- Wired into `StrengthSessionRunner`: "Elapsed" before an exercise's first set, "Rest" after, timing from that set's `completedAt`. 2 new component smoke tests.
- `deriveRestIntervals` (post-hoc rest between logged sets, for S1.7/future rest-aware cost modelling) is separate from the live countdown for a structural reason: the live display counts up *before* a "next" set exists, so there's no second timestamp to derive an interval from yet.

## Validation

- [x] `cd app && npm run check` — pass: 1305 tests, 108 files.

Results:
- `elapsedSeconds`/`formatElapsed`/`deriveRestSecondsBetweenSets`/`deriveRestIntervals` fully covered (11 tests, including flooring, negative-clock-skew guard, unparseable-instant guard).
- [ ] Not verified: "backgrounding for two minutes shows correct elapsed rest" and "no timer state survives reload" are real browser behaviour a Node test cannot exercise. Structurally argued instead: the hook's only state is `now`, reset fresh on mount, and the rendered value is a pure function of `sinceIso` (Firestore-durable, S1.1) and wall-clock time — there is no accumulator to lose.

## Risk and reviewer guidance

Small, low-risk PR. The one thing worth a second look: the timestamp-not-accumulator design is the entire point of this item (backgrounded-tab timer throttling is routine on mobile) — worth confirming the reasoning holds rather than just the diff.

## Domain invariants

- [x] Firestore writes remain user-scoped — no new writes; reads `completedAt`/`startedAt` already persisted by S1.4/S1.5.
- [x] Calendar-date logic uses Europe/Warsaw — not applicable; this is elapsed-*time* (seconds), not calendar-date logic.
- [x] Recommendation decision logic changed: no.

## Screenshots or recordings

Not captured — no live browser session available in this environment. Recommend a manual pass on a real device to confirm the timer survives backgrounding correctly.

### DIFF
--- a/app/src/components/StrengthSessionRunner.css
+++ b/app/src/components/StrengthSessionRunner.css
@@ -146,11 +146,27 @@
   color: var(--text-secondary, #94a3b8);
 }
 
-.set-gauge { margin-left: auto; }
-.set-sync { margin-left: 0.35rem; font-size: 0.72rem; white-space: nowrap; }
-.set-sync-synced { color: var(--success, #22c55e); }
-.set-sync-pending { color: var(--warning, #f59e0b); }
-.set-sync-unavailable { color: var(--danger, #ef4444); }
+.set-gauge {
+  margin-left: auto;
+}
+
+.set-sync {
+  margin-left: 0.35rem;
+  font-size: 0.72rem;
+  white-space: nowrap;
+}
+
+.set-sync-synced {
+  color: var(--success, #22c55e);
+}
+
+.set-sync-pending {
+  color: var(--warning, #f59e0b);
+}
+
+.set-sync-unavailable {
+  color: var(--danger, #ef4444);
+}
 
 .strength-set-entry {
   display: grid;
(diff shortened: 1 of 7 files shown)


Szczepanov:
S1.6 review complete and improved in commit a4c8cf2. Rest timing now follows the latest completed set across the entire session, so switching exercises or alternating a superset no longer resets the timer to session start. Regression coverage was added.

PR #77, opened 2026-08-17
feat(strength): progressive-overload history view (S1.7)

Szczepanov:
## Summary

- Closes Step A. `workouts/overloadHistory.ts`: `summarizeExerciseSession`, `summarizeExerciseAcrossSessions`, `distinctLoggedExercises` — pure, 14 tests, reading the raw per-set log directly (D-SETLOG). Nothing here depends on S2.1's 1RM estimator or Step C, neither of which exist yet at this point in the plan.
- **A real identity bug was caught and fixed during implementation, not after**: the first draft matched an exercise across sessions on `exerciseId` alone. Two different free-text lifts both have `exerciseId: null`, so that match would have silently merged e.g. "Farmer carry" and "Sled push" into one shared history. Every lookup now goes through an `ExerciseIdentity{exerciseId, freeTextName}` and one `identityKey` function; a dedicated test pins the fix.
- "Best set" is a plain heaviest-weight-moved heuristic, not an estimate — never conflated with S2.1's (not-yet-built) 1RM estimator, which uses a different, gauge-filtered admissibility rule. Named `heaviestSet` in the type itself, not "best" or "estimated max", to make that boundary hard to blur later by accident.
- `strengthSessionService.getSessionsInRange`: bounded date-range read, same query shape as `activityService.getActivitiesInRange`/`decisionJournalService.getEntriesInRange` — an invalid document is omitted from the rows and counted, not allowed to fail the whole range. 3 new tests.
- `components/StrengthOverloadHistory.tsx`: a table, not a chart — the plan's own wording ("load, reps, tonnage and best set over time") is satisfied by one row per session. 8 smoke tests. Wired into the same `'strength'` screen as S1.5's runner, stacked below it.

## Validation

- [x] `cd app && npm run check` — pass: 1330 tests, 110 files.

Results:
- 14 tests for the aggregation module, including the free-text-conflation regression test; 3 for the new service query; 8 component smoke tests.

## Risk and reviewer guidance

The free-text identity bug (see Summary) is the thing worth double-checking most closely — it's exactly the class of bug that's easy to miss on review since the "obviously correct" version (match on `exerciseId`) looks fine until two null-id exercises coexist.

## Domain invariants

- [x] Firestore writes remain user-scoped — `getSessionsInRange` reads only `users/{userId}/strength_sessions`.
- [x] Calendar-date logic uses Europe/Warsaw — history sorted/queried on the already-Warsaw-local `date` field from S1.4; no new date computation.
- [x] Recommendation decision logic changed: no. Read-only history view.

## Screenshots or recordings

Not captured — no live browser session available in this environment. Recommend a manual pass to confirm the table renders correctly for full/partial/no telemetry.

### DIFF
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -523,6 +523,11 @@
   }
 }
 
+.strength-screen {
+  display: grid;
+  gap: 1rem;
+}
+
 /* Smooth transitions */
 .app-content > * {
   animation: fadeIn 0.3s ease;
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -9,6 +9,7 @@ import { Preferences } from './components/Preferences';
 import { DataView } from './components/DataView';
 import { ExternalPlanImport } from './components/ExternalPlanImport';
 import { StrengthSessionRunner } from './components/StrengthSessionRunner';
+import { StrengthOverloadHistory } from './components/StrengthOverloadHistory';
 import { decisionComposer } from './engine/composer';
 import type { DailyDecisionInput } from './engine/models';
 import type { Screen } from './types/navigation';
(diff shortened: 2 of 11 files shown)


Szczepanov:
S1.7 review complete and improved in commit a6ae223. The history query now uses an exact exclusive 90-day window; exercise identity is normalized/namespaced; repeated exercise entries are aggregated; and same-day ordering is deterministic.

PR #78, opened 2026-08-17
feat(strength): gauge-aware 1RM estimator (S2.1)

Szczepanov:
## Summary

- `workouts/oneRepMax.ts`: `estimateOneRepMax`, pure, no I/O, 11 tests. Epley (1985) formula, stated with why in the module docstring: simplicity and being the more common default, explicitly **not** a claim it outperforms Brzycki — no evidence either way here, and per D-SUBJEST's precedent (ADR-0020) the formula is versioned policy, not an invariant this module asserts authority over.
- The gauge filter is the point of this module, each rule independently tested: `isWarmup` excluded; `null` `weightKg` (bodyweight) excluded, no external load to estimate a loaded max from; reps beyond `MAX_REPS_FOR_ESTIMATION` (12) excluded even near failure, since Epley's error compounds with rep count regardless; `velocity_loss`/`technical` gauges excluded unconditionally at any value — quality-limited sets imply nothing about a true maximum; an absent gauge excluded by default, the explicit deliberate choice the plan asked for; `rir`/`rpe_rts` admissible only within `MAX_RIR_FOR_ESTIMATION` (3), i.e. RPE ≥ 7 — the source analysis's own finding on `primer_rir`'s default of 6, where a self-estimate that far out is a coaching instruction, not a measurement.
- Among admissible sets, returns whichever implies the highest capacity, not an average — one genuine near-failure set is stronger evidence than diluting it against other admissible sets.

## Validation

- [x] `cd app && npm run check` — pass: 1341 tests, 111 files.

Results:
- 11 tests covering every admissibility rule independently (warm-up, bodyweight, rep cap, each gauge scale, RIR/RPE thresholds, empty input, highest-capacity selection).

## Risk and reviewer guidance

The admissibility thresholds (`MAX_RIR_FOR_ESTIMATION = 3`, `MAX_REPS_FOR_ESTIMATION = 12`) are this implementation's own documented choices, not derived from a cited source — worth a second opinion if you have a stronger reference for either.

## Domain invariants

- [x] Firestore writes remain user-scoped — no writes in this PR; pure estimation function only (write-back is S2.2, next PR).
- [x] Calendar-date logic uses Europe/Warsaw — not applicable, no date logic here.
- [x] Recommendation decision logic changed: no. Estimator has no caller yet; nothing consumes its output until S2.2.

## Screenshots or recordings

### DIFF
--- a/app/src/workouts/oneRepMax.test.ts
+++ b/app/src/workouts/oneRepMax.test.ts
@@ -37,6 +37,11 @@ describe('estimateOneRepMax', () => {
         expect(estimateOneRepMax([set({ gauge: undefined })])).toBeNull();
     });
 
+    it('excludes an out-of-range gauge rather than treating it as evidence', () => {
+        expect(estimateOneRepMax([set({ gauge: { scale: 'rir', value: -1 } })])).toBeNull();
+        expect(estimateOneRepMax([set({ gauge: { scale: 'rpe_rts', value: 11 } })])).toBeNull();
+    });
+
     it(`excludes a set beyond MAX_RIR_FOR_ESTIMATION (${MAX_RIR_FOR_ESTIMATION}) reps in reserve`, () => {
         expect(estimateOneRepMax([set({ gauge: { scale: 'rir', value: MAX_RIR_FOR_ESTIMATION + 1 } })])).toBeNull();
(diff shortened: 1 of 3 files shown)


Szczepanov:
S2.1 review complete and improved in commit d77ef86. Out-of-range or otherwise invalid gauges can no longer qualify as near-failure evidence for a 1RM estimate. Added explicit invalid-RIR/RPE regressions.

PR #79, opened 2026-08-17
feat(strength): 1RM write-back under a derived ownership rung (S2.2)

Szczepanov:
## Summary

- Closes Step B. `workouts/oneRepMaxWriteback.ts`: `applyOneRepMaxDerivation` (single exercise), `deriveOneRepMaxUpdatesForSession` (whole session) — pure, 9 tests. `preferencesService.applyOneRepMaxDerivations` wires it to Firestore, 4 tests against a mocked SDK.
- `workouts/models.ts`: `TargetSource` widened to add `'derived'`, reused by both `targetSources` and the new `estimated1RmSources` map (per-exercise provenance, parallel to `estimated1RmKg`). `engine/validation.ts` validates the new field the same way `estimated1RmKg` already is.
- **A first draft contradicted the accepted ADR-0021 text and was caught before landing.** The plan's own wording ("existing source is absent or itself derived") is ambiguous; the first implementation resolved that by treating an existing value with no recorded source as *protected*, by analogy with `upsert_garmin_performance_targets`' "no provenance → treat as manual" precedent. Re-reading ADR-0021 directly before writing the outcome note surfaced its actual unambiguous text: *"A derived estimate may populate a target whose source is absent or already derived."* Opposite of what was built. Fixed before commit: writability now turns on `existingSource` alone (`undefined` or `'derived'`), never on whether a value already exists under that absent source. One test changed to match.
- Dual-write: `prescription.ts` reads `strength.estimated1RmKg` ahead of the legacy top-level `estimated1RmKg`, so `applyOneRepMaxDerivations` reads the same effective value and writes the result to both, keeping neither reader stale. `estimated1RmSources` stays a single shared provenance map — there is exactly one true source per exercise.
- `prescription.test.ts`'s required "relative when no 1RM is known" case run explicitly and confirmed passing, per the plan's stated acceptance criterion.

## Validation

- [x] `cd app && npm run check` — pass: 1354 tests, 113 files.

Results:
- 9 tests for the pure write-back rule; 4 for the Firestore-wired service method (including a legacy-vs-`strength.*` precedence case); `prescription.test.ts`'s "relative when no 1RM is known" test run and confirmed still green.

## Risk and reviewer guidance

**Please read the ADR-contradiction note in the Summary carefully** — it's the most important thing in this PR. I want a second set of eyes confirming my re-read of ADR-0021 (an absent source is writable, full stop, regardless of whether a value already exists) is correct, since it directly affects whether an untracked-provenance 1RM value can be silently overwritten by a derived estimate.

## Domain invariants

- [x] Firestore writes remain user-scoped — writes go through `preferencesService`, existing user-scoped path (`users/{userId}/preferences/profile`).
- [x] Calendar-date logic uses Europe/Warsaw — not applicable, no date logic in this PR.
- [x] Recommendation decision logic changed: no `POLICY_VERSION` impact — this changes prescription *input data* (1RM), not the recommendation engine's decision logic itself, and only for exercises with no `manual`/`coach` source already set.

## Screenshots or recordings

### DIFF
--- a/app/src/engine/validation.ts
+++ b/app/src/engine/validation.ts
@@ -873,6 +873,19 @@ export function validatePreferences(raw: any): ValidationResult<UserPreferences>
                  !Object.values(raw.performanceProfile.estimated1RmKg).every((value) => typeof value === 'number' && Number.isFinite(value) && value > 0))) {
                 errors.push({ field: 'performanceProfile.estimated1RmKg', message: 'All estimated 1RM values must be positive numbers' });
             }
+            // ADR-0021 D-1RMSRC: per-exercise provenance for estimated1RmKg.
+            if (raw.performanceProfile.estimated1RmSources !== undefined) {
+                const validSources = ['garmin', 'manual', 'coach', 'derived'];
+                const sources = raw.performanceProfile.estimated1RmSources;
+                const isValid = typeof sources === 'object' && sources !== null &&
+                    Object.values(sources).every((entry: any) =>
+                        typeof entry === 'object' && entry !== null &&
+                        validSources.includes(entry.source) &&
+                        (entry.computedAt === undefined || typeof entry.computedAt === 'string'));
+                if (!isValid) {
+                    errors.push({ field: 'performanceProfile.estimated1RmSources', message: 'Each 1RM source entry must have a valid source (garmin, manual, coach or derived) and an optional computedAt string' });
+                }
+            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


Szczepanov:
S2.2 review complete and improved in commit 8b9d2a6. The write-back result is now a discriminated union without an unsafe protected-value cast. Production session completion derives/writes eligible 1RM values before closing the session, so a failed write leaves the workout resumable instead of terminal and partially processed.

PR #80, opened 2026-08-17
feat(strength): set log -> CompletedExposure derivation (S3.1)

Szczepanov:
## Summary

- **Builds default-off**: has no effect on any recommendation until S3.2 (next PR) wires `manualTraining` in behind a selector, and S3.3 measures it. Permitted by ADR-0021's D-STRCOST, which approves *building and measuring*, not prescribing new coefficients — this item stays inside that boundary by reusing the existing, already-in-production `DEFAULT_COST_BY_MODALITY.Strength` / `DEFAULT_STIMULUS_BY_MODALITY.Strength` tables rather than asserting new magnitudes. What's new is a redistribution of the flat table's `lowerBody`/`upperBody` split by which exercises were actually logged (`primaryMuscles`, weighted by working-set count), and a `maxStrength`/`hypertrophy`-only stimulus per the plan's own scoping.
- `workouts/strengthExposure.ts`: `deriveStrengthExposure`, pure, 17 tests.
- **Two real defects were caught by this module's own tests before this commit**:
  1. *Architecture defect*: the first draft summed a full session-sized budget once per exercise, so one hard exercise and five hard exercises became indistinguishable (both saturated at the same capped value). Redesigned to look up the session's budget once, at the hardest exercise present, and only weight-redistribute that one budget's split across exercises. The failing test is now a permanent regression guard.
  2. *Bound violation*: `WorkoutCostProfile` axes are canonically 0.0–1.0. A fully one-sided exercise (`front_squat`, no upper-body muscle at all) at `hard` intensity redistributed the combined 1.5 budget entirely onto `lowerBody`, exceeding 1.0 — caught by a test, fixed with a per-axis cap.
- Evidence tier is the weakest link, not the best: one free-text exercise among several identified ones drags the whole session to `unknown` confidence, not partial credit.

## Validation

- [x] `cd app && npm run check` — pass: 1371 tests, 114 files.

Results:
- 17 tests, including the two regression guards for the defects above and explicit gauge/free-text/idempotency cases.

## Risk and reviewer guidance

Read the two "caught by tests" defects in the Summary — they're the interesting part of this PR, showing the design actually converge through testing rather than being right on the first pass. This PR has **zero effect on any recommendation** as written (nothing calls `deriveStrengthExposure` yet); the risk surface is entirely "is this the right derivation for when it does get wired in," which S3.2/S3.3 gate behind measurement anyway.

## Domain invariants

- [x] Firestore writes remain user-scoped — no writes in this PR, pure derivation function only.
- [x] Calendar-date logic uses Europe/Warsaw — session `date` (already Warsaw-local from S1.4) is passed through unchanged; duration computed from `startedAt`/`completedAt` instants, not calendar dates.
- [x] Recommendation decision logic changed: no. Function exists but has no caller; `POLICY_VERSION` untouched.

## Screenshots or recordings

### DIFF
--- a/app/src/workouts/oneRepMax.ts
+++ b/app/src/workouts/oneRepMax.ts
@@ -30,7 +30,10 @@ export const MAX_RIR_FOR_ESTIMATION = 3;
  *  documented cap rather than an unbounded formula applied everywhere it technically runs. */
 export const MAX_REPS_FOR_ESTIMATION = 12;
 
-function isNearFailureGauge(gauge: IntensityGauge | undefined): boolean {
+/** Exported so S3.1's exposure derivation can classify per-exercise intensity using the
+ *  same "how close to failure" definition, rather than a second, potentially-drifting copy
+ *  of this rule. */
+export function isNearFailureGauge(gauge: IntensityGauge | undefined): boolean {
     if (!gauge) return false; // No gauge at all: excluded by default, not a guessed fallback (ADR-0021 S2.1).
     switch (gauge.scale) {
         case 'rir':
--- a/app/src/workouts/strengthExposure.test.ts
+++ b/app/src/workouts/strengthExposure.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, it } from 'vitest';
+import { deriveStrengthExposure } from './strengthExposure';
+import type { LoggedSet, StrengthSession } from '../engine/models';
+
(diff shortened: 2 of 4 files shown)


Szczepanov:
S3.1 review complete and improved in commit 5e917da. Non-strength catalog IDs no longer masquerade as identified strength work; evidence confidence uses athlete classification consistently; and occurrence keys preserve recommendation/date identity to support downstream deduplication.

PR #81, opened 2026-08-17
feat(engine): manualTraining source wiring behind a default-off selector (S3.2)

Szczepanov:
## Summary

- `ManualTrainingPolicy = 'off' | 'included'` added to `trainingHistorySnapshot.ts`, mirroring `SubjectiveDriftPolicy`'s plumbing exactly (`rules.ts`, ADR-0020) rather than inventing a second selector shape. `firestoreTrainingHistory.ts` (the one production call site) now fetches real strength sessions in parallel with activities/recommendations, but calls the builder with the policy literal `'off'` — the fetch is wired so enabling the source later is a policy flip, not a second fetch-wiring change, but it currently has zero effect.
- **Every branch checks `=== 'off'`, never `=== 'included'`.** A first draft did the opposite and a new architecture guard test — mirroring `rules.test.ts`'s existing `SubjectiveDriftPolicy` guard exactly — caught it immediately: the enabled literal must never appear in production engine code outside a test or a future `engine/simulation/` harness. Fixed by inverting every branch, matching how `rules.ts`'s `subjectiveDriftStrain` never once writes the literal `'drift'` outside its own type declaration.
- The revision string's *shape*, not only its value, had to stay identical when off — an early version always appended a fifth segment; caught before commit.
- A second failure-contract edge case: zero usable sessions with a nonzero `invalidRecords` count (every document present was corrupt) must report `INVALID`, not `MISSING`. Extracted into an exported, directly-testable `toManualTrainingDataState` specifically because the selector being off in production means `getSnapshot`'s returned snapshot cannot observably distinguish the two states.
- **`simulate:diff` could not be run to a clean pass, for a reason unrelated to this change.** Verified pre-existing (not introduced here) by running the identical command against an unmodified `main` checkout in a disposable git worktree — byte-identical output. This stack's own diff contribution is confirmable as zero: `git diff main...HEAD` touches no file under `engine/simulation/` or the baseline JSON. Not fixed here — regenerating the baseline requires human review per this repo's own tooling and is unrelated to strength logging.

## Validation

- [x] `cd app && npm run check` — pass: 1384 tests, 115 files.
- [x] `cd app && npm run simulate:scenarios` — n/a, ran `simulate:diff` instead per plan's stated criterion; see note above (pre-existing, unrelated staleness, verified via worktree comparison against clean `main`).

Results:
- 9 new tests in `trainingHistorySnapshot.test.ts` (including the architecture guard), 6 in a new `firestoreTrainingHistory.test.ts`.

## Risk and reviewer guidance

**Please independently verify the `simulate:diff` pre-existing-staleness claim** — it's the one thing in this stack I couldn't get to a clean "zero diff" state, and I want a second check that the baseline mismatch really does predate this work rather than being masked by something in my verification method.

## Domain invariants

- [x] Firestore writes remain user-scoped — read-only fetch added (`strengthSessionService.getSessionsInRange`), user-scoped path unchanged.
- [x] Calendar-date logic uses Europe/Warsaw — date range computed via existing `addDaysToLocalDateString`, unchanged pattern.
- [x] Recommendation decision logic changed: **no, by design** — selector defaults to and stays `'off'` at the only production call site; `POLICY_VERSION` untouched. Architecture guard test enforces this cannot regress silently.

## Screenshots or recordings

### DIFF
--- a/app/src/engine/firestoreTrainingHistory.test.ts
+++ b/app/src/engine/firestoreTrainingHistory.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it, vi } from 'vitest';
+import type { StrengthSession } from './models';
+
+const { activityService, recommendationService, strengthSessionService } = vi.hoisted(() => ({
+    activityService: { getActivitiesInRange: vi.fn(async () => ({ status: 'AVAILABLE', revision: 'a', data: [] })) },
+    recommendationService: { getRecommendationsInRange: vi.fn(async () => ({ status: 'AVAILABLE', revision: 'r', data: [] })) },
+    strengthSessionService: { getSessionsInRange: vi.fn(async (): Promise<{ sessions: StrengthSession[]; invalidRecords: number }> => ({ sessions: [], invalidRecords: 0 })) },
+}));
+
+vi.mock('../services/activityService', () => ({ activityService }));
+vi.mock('../services/recommendationService', () => ({ recommendationService }));
+vi.mock('../services/strengthSessionService', () => ({ strengthSessionService }));
+
+import { firestoreTrainingHistoryProvider, toManualTrainingDataState } from './firestoreTrainingHistory';
+
+const session: StrengthSession = {
+    userId: 'u1', sessionId: 's1', date: '2026-08-06', startedAt: '2026-08-06T18:00:00Z',
+    updatedAt: '2026-08-06T18:45:00Z', state: 'completed', exercises: [], schemaVersion: 1,
+};
+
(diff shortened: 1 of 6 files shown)


Szczepanov:
S3.2 review complete and improved in commits fb6a5d9 and d109f4e. Default-off mode performs no strength-session read and preserves the historical snapshot shape. Enabled mode uses an exclusive bounded read, treats any corrupt record as INVALID, maps read failures to UNAVAILABLE/blocking, revisions by session update, and replaces matching Garmin/adherence occurrences instead of double-counting them. The production-off policy is now documented.

PR #82, opened 2026-08-17
docs(strength): S3.3 measurement infrastructure and defer decision

Szczepanov:
## Summary

- Closes the strength-session-logging plan's task board (all 12 items). The honest constraint driving this item: every item in Steps A–C was built and landed in this single stack, so **zero real strength sessions exist to measure against**. This is explicitly not the same outcome as D-BEAM's "measured, no improvement" precedent — it's "cannot yet be measured at all" — and the plan's outcome note is explicit about that distinction rather than blurring it.
- `workouts/manualTrainingMeasurement.ts`: `summarizeManualTrainingMeasurement`, pure, 4 tests — aggregates cost/stimulus/confidence across a set of sessions, independent of the selector so it can run at any time without enabling anything. Deliberately scoped smaller than a full per-day recommendation-verdict comparison generator (the `simulate:diff`/`subjectiveDriftComparison.ts` shape D-BEAM's Phase 9.6 precedent used): building that now, with nothing real to run it against, would be exactly the premature investment this plan's cheapest-risk-first sequencing exists to defer.
- **Decision recorded: DEFER.** `ManualTrainingPolicy` stays `'off'` in production. Fabricating synthetic session data to produce a report with real-looking numbers was explicitly rejected — per ADR-0020's D-SUBJCAL precedent, synthetic-only evidence does not authorize shipping a measured candidate, and a fabricated report could be mistaken for real evidence, which is worse than reporting none. The plan records what "ready to measure" looks like (a real training block's worth of logged sessions) so this stays actionable rather than indefinite.
- Plan header's stale "Blocked by" line corrected — D-STRCOST is accepted and Steps A–C are built; only *enabling* the source in production remains deferred.

## Validation

- [x] `cd app && npm run check` — pass: 1388 tests, 116 files.

## Risk and reviewer guidance

Zero code-behaviour risk — this PR is measurement infrastructure plus a documentation decision record, and the production selector stays off. The thing worth reading closely is the *decision itself*: confirm you agree "defer, don't fabricate evidence" is the right call rather than either shipping without real data or building a much larger synthetic-scenario harness speculatively.

## Domain invariants

- [x] Firestore writes remain user-scoped — no writes in this PR.
- [x] Calendar-date logic uses Europe/Warsaw — not applicable, no date logic in this PR.
- [x] Recommendation decision logic changed: no. `ManualTrainingPolicy` stays `'off'`; `POLICY_VERSION` untouched.

## Screenshots or recordings

### DIFF
--- a/app/src/workouts/manualTrainingMeasurement.test.ts
+++ b/app/src/workouts/manualTrainingMeasurement.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest';
+import { summarizeManualTrainingMeasurement } from './manualTrainingMeasurement';
+import type { StrengthSession } from '../engine/models';
+
+function session(overrides: Partial<StrengthSession> = {}): StrengthSession {
+    return {
+        userId: 'u1', sessionId: 's1', date: '2026-08-17', startedAt: '2026-08-17T18:00:00Z',
+        completedAt: '2026-08-17T19:00:00Z', updatedAt: '2026-08-17T19:00:00Z', state: 'completed',
+        exercises: [], schemaVersion: 1,
+        ...overrides,
+    };
+}
+
+describe('summarizeManualTrainingMeasurement', () => {
+    it('reports an empty summary for no sessions -- the actual state today', () => {
+        const summary = summarizeManualTrainingMeasurement([]);
+        expect(summary).toEqual({
+            sessionCount: 0, exposureCount: 0, skippedSessionCount: 0,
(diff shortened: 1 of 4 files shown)


Szczepanov:
S3.3 review complete and improved in commit b4d6d20. Measurement now deduplicates physical occurrences and reports duplicate counts. The DEFER decision remains correct because no real athlete dataset supports activation yet; production remains default-off. Consolidated validation on this head passes: build/typecheck/lint, 1,400 unit tests, 58 Firestore emulator tests, and workout-catalog validation.

PR #96, opened 2026-08-19
feat(sessions): fine-grained v2 import preview and revision diff (M3.7)

Szczepanov:
Closes M3.7's two remaining gaps from `docs/plans/multidomain-session-authoring-execution-and-evidence.md`:

- `ExternalPlanImport` now renders the full `SessionDefinitionPreview` (blocks/steps/dose/effort/option sets) for each `external-plan@2` session, not just a one-line summary.
- New `sessions/sessionDefinitionDiff.ts` computes a block/step/choice-level revision diff between two `SessionDefinition`s, tagging every row `behaviorChanging` or cosmetic. `externalPlanDiff.ts` uses it for v2/v2 session pairs, replacing the coarse "content changed" line; v1 keeps its coarse check since `ExternalPrescription` has no comparable structure.
- The import preview lists diff rows inline (⚠ for behavior-changing) and requires an explicit acknowledgement checkbox before Import enables whenever unreviewed behavior changes exist.

Verified with `npm run check` (typecheck, lint, unit tests, catalog validation) — full plan outcome notes are in the doc.

## Summary by CodeRabbit

* **New Features**
  * Added detailed previews of v2 session definitions during external plan imports.
  * Added fine-grained highlighting for session, block, step, choice, and option changes.
  * Distinguished behavior-changing updates from cosmetic text changes.
  * Added expandable session details and clearer visual change indicators.
  * Import confirmation now requires acknowledgment of behavior-changing updates.

* **Documentation**
  * Documented completion of full session-content previews and behavior-change review requirements.

### DIFF
--- a/app/src/sessions/sessionDefinitionDiff.ts
+++ b/app/src/sessions/sessionDefinitionDiff.ts
@@ -1,0 +1,316 @@
+/**
+ * M3.7: fine-grained content diff between two revisions of the same `SessionDefinition`.
+ *
+ * `externalPlanDiff.ts` already flags placement/gating/objective/scaling changes at the
+ * plan-session envelope level; this module is the "the session content changed" line's
+ * replacement for `external-plan@2` sessions (and any other two `SessionDefinition`
+ * revisions worth comparing) -- it names *what* changed inside `blocks`, not just that
+ * something did.
+ *
+ * Every row is tagged `behaviorChanging`. That distinguishes changes that alter what the
+ * athlete is asked to do or how the runner behaves (dose, load, laterality, optional vs
+ * required, an authored choice's actions, ...) from purely cosmetic text (title, notes,
+ * a choice's trigger wording). The preview surfaces both, but only behavior-changing rows
+ * need to block casual confirmation.
+ */
+import type {
+    RangeOrNumber,
+    SessionBlock,
+    SessionChoice,
+    SessionChoiceAction,
+    SessionDefinition,
+    SessionDose,
+    SessionEffort,
+    SessionLoad,
+    SessionOption,
+    SessionQuality,
+    SessionStep,
+} from './models';
+
+export interface SessionContentDiffRow {
+    scope: 'session' | 'block' | 'step' | 'choice' | 'option';
+    id: string;
+    change: 'added' | 'removed' | 'changed';
+    behaviorChanging: boolean;
+    detail: string;
+}
+
+function formatRangeOrNumber(value: RangeOrNumber): string {
+    return typeof value === 'number' ? String(value) : `${value.min}–${value.max}`;
+}
+
+function sameJson(left: unknown, right: unknown): boolean {
+    return JSON.stringify(left ?? null) === JSON.stringify(right ?? null);
+}
+
+function sameStringSet(left: readonly string[] | undefined, right: readonly string[] | undefined): boolean {
+    return JSON.stringify([...(left ?? [])].sort()) === JSON.stringify([...(right ?? [])].sort());
+}
+
+function formatDose(dose: SessionDose | undefined): string {
+    if (!dose) return 'no prescribed dose';
+    switch (dose.kind) {
+        case 'repetition': return `${dose.sets} × ${formatRangeOrNumber(dose.reps)} reps`;
+        case 'duration': return `${dose.sets ?? 1} × ${formatRangeOrNumber(dose.seconds)} s`;
+        case 'distance': return `${formatRangeOrNumber(dose.meters ?? dose.metres ?? 0)} m`;
+        case 'checkoff': return `${dose.rounds ?? 1} rounds`;
+    }
+}
+
+function formatLoad(load: SessionLoad | undefined): string {
+    if (!load) return 'no prescribed load';
+    switch (load.kind) {
+        case 'bodyweight': return 'bodyweight';
+        case 'mass': return `${formatRangeOrNumber(load.kg)} kg`;
+        case 'band': return `${load.bandColor} band`;
+        case 'percent_max': return `${load.percent}% max`;
+        case 'percent_one_rm': return `${load.percent}% 1RM`;
+        case 'relative_step': return `${load.percent}% of ${load.stepId}`;
+        case 'descriptive': return load.display ?? load.description ?? 'descriptive load';
+        case 'unloaded': return 'unloaded';
+    }
+}
+
+function formatEffort(effort: SessionEffort | undefined): string {
+    if (!effort) return 'none';
+    const parts: string[] = [];
+    if (effort.target !== undefined) parts.push(`${effort.kind ?? 'target'} ${formatRangeOrNumber(effort.target)}`);
+    if (effort.rpe !== undefined) parts.push(`RPE ${formatRangeOrNumber(effort.rpe)}`);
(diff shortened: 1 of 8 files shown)


app/src/sessions/sessionDefinitionDiff.ts
```diff
@@ -0,0 +1,336 @@
...
+        case 'band': return `${load.bandColor} band`;
+        case 'percent_max': return `${load.percent}% max`;
+        case 'percent_one_rm': return `${load.percent}% 1RM`;
+        case 'relative_step': return `${load.percent}% of ${load.stepId}`;
+        case 'descriptive': return load.display ?? load.description ?? 'descriptive load';
+        case 'unloaded': return 'unloaded';
+    }
+}
+
+function formatEffort(effort: SessionEffort | undefined): string {
```
Szczepanov, line 93: `formatAction()` never renders `targetStepId`/`targetBlockId`, so an authored choice's action that gets re-targeted to a different step/block prints identical before/after text.

Example: an option's actions change from `{kind:'reduce_load_percent', targetStepId:'step-squat', percent:10}` to `{kind:'reduce_load_percent', targetStepId:'step-deadlift', percent:10}` — a real behavior change (load reduction moves to a different exercise). `sameJson` correctly flags `behaviorChanging:true`, but the rendered detail reads "actions reduce load 10% → reduce load 10%", giving the athlete no information about what actually changed while still forcing the acknowledgment checkbox for an apparently no-op change.

**Fixed** in a follow-up commit: each action kind now includes its target step/block id in the formatted string.

---

app/src/sessions/sessionDefinitionDiff.ts
```diff
@@ -0,0 +1,336 @@
...
+    change: 'added' | 'removed' | 'changed';
+    behaviorChanging: boolean;
+    detail: string;
+}
+
+function formatRangeOrNumber(value: RangeOrNumber): string {
+    return typeof value === 'number' ? String(value) : `${value.min}–${value.max}`;
+}
+
+/** Recursively sorts object keys so structurally-identical values compare equal regardless
```
Szczepanov, line 42: `sameJson()` uses plain `JSON.stringify` for structural equality, so two semantically identical dose/load/effort/quality/alternatives objects with differently-ordered keys are reported as changed.

This isn't just theoretical here: `ExternalPlanImport.tsx`'s own workflow asks an LLM to regenerate the full plan JSON on each revision. Two independent generations of an unchanged step commonly produce `{"kind":"repetition","sets":4,"reps":5}` vs `{"kind":"repetition","reps":5,"sets":4}` — `JSON.stringify` differs, so `diffStep` reports a phantom `behaviorChanging` row, forcing an acknowledgment for nothing and eroding trust in real warnings.

**Fixed**: `sameJson` now canonicalizes (recursively sorts object keys) before comparing.

---

app/src/sessions/sessionDefinitionDiff.ts
```diff
@@ -0,0 +1,336 @@
...
+    if (Array.isArray(value)) return value.map(canonicalize);
+    if (value !== null && typeof value === 'object') {
+        const sorted: Record<string, unknown> = {};
+        for (const key of Object.keys(value as Record<string, unknown>).sort()) {
+            sorted[key] = canonicalize((value as Record<string, unknown>)[key]);
+        }
+        return sorted;
+    }
+    return value;
+}
```
Szczepanov, line 55: `formatDose('distance')` falls back to `0` when both `meters` and `metres` are undefined, displaying a fabricated "0 m" instead of indicating no value is set. `validation.ts`'s distance-dose check allows this (`required=false`), so it's reachable: if such a step's `sets` count changes, the preview prints "dose 0 m → 0 m", which reads as a concrete prescribed distance that never existed.

**Fixed**: now renders "no distance set" when neither field is present.

---

app/src/sessions/sessionDefinitionDiff.ts
```diff
@@ -0,0 +1,336 @@
...
+
+function diffStep(blockId: string, before: SessionStep, after: SessionStep): SessionContentDiffRow | null {
+    const changes: string[] = [];
+    let behaviorChanging = false;
+
+    if (!sameJson(before.exerciseRef, after.exerciseRef)) {
+        changes.push('movement changed');
+        behaviorChanging = true;
+    }
```
Szczepanov, line 208: The rest-change message unconditionally appends `' s'` even when the value renders as `'none'`, e.g. `before.rest=60, after.rest=undefined` produces "rest 60 → none s" — the unit suffix incorrectly binds to "none" rather than the numeric value.

**Fixed**: the unit is now attached per-side, only when that side has a numeric value.

---

Szczepanov (commented):
Reviewed `sessionDefinitionDiff.ts` (the new revision-diff engine). Found four correctness issues in the diff-rendering logic — all fixed in commit a30f709, pushed to this branch:

1. `formatAction()` didn't render action targets, hiding real re-targeting changes behind identical before/after text.
2. `sameJson()` used raw `JSON.stringify`, so key-order differences (plausible given the LLM-regenerated import JSON) produced phantom behavior-changing rows.
3. `formatDose('distance')` fabricated "0 m" for an unset distance.
4. The rest-change message misattached its ` s` unit suffix to the literal "none".

---

Szczepanov:
Addressed the remaining CodeRabbit findings not already covered by commit a30f709, in a30f709's follow-up (82cc66d):

- **Distance dose omitted `sets`** — `formatDose('distance')` now renders `${sets} × ...`, matching the repetition/duration format, so a sets-only change is no longer invisible in the diff text.
- **Block/step reordering undetected** — the `byId()`-keyed comparison is deliberately order-blind (diffs by identity, not position), so swapping two existing blocks/steps produced no diff row and could bypass import acknowledgement. Added `orderChanged()`: a same-membership, different-position check layered on top of the existing add/remove detection.
- **Test-coverage claim accuracy** — the plan doc's verification paragraph claimed both directions of laterality change and both add+remove for blocks/steps, but only tested one direction/case each. Added the missing reverse-direction laterality test, an added-step+removed-block test (complementing the existing removed-step+added-block one), a reordered-object-keys test, and new cases for the two fixes above.
- **`ExternalPlanImport.test.ts` interaction coverage (nitpick)** — exported `PlanPreview` so its acknowledgement gating is directly testable, and added cases asserting Import renders disabled+checkbox-shown for a behavior-changing diff and enabled+no-checkbox for a cosmetic-only one. This repo has no interactive component-test harness (no `@testing-library/react`), so the "clicking the checkbox re-enables Import" half isn't independently coverable — the underlying state is a plain `useState` toggle with no branching logic of its own to hide a bug in. Renamed the file to `.test.tsx` since it now contains JSX.

**Skipped with reasoning:** the outside-diff suggestion to strike M3.7's `Progress`/`Change`/`Done when` paragraphs now that it's complete. Every other completed milestone in this same plan doc (M3.1–M3.6) keeps its original spec text alongside an appended `Outcome` section — that's the doc's established convention throughout, not something to break in one section of one PR. Happy to do a doc-wide pass separately if wanted.

`sameJson()`'s key-order issue (CodeRabbit's 4th actionable item) was already fixed in a30f709 (`canonicalize()`), so no further change needed there.

Rebased/merged this fix through the whole PR stack (#97–#100) so each PR's diff stays clean against its moving base. `npm run check` + the Firestore emulator suite (82/82) pass at every point in the chain.

PR #97, opened 2026-08-19
feat(sessions): bounded M3.8 builder hardening (load/effort/choices)

Szczepanov:
Stacked on #96 (M3.7).

Adds the two gaps the M3.8 cutline named, then stops per its own "do not build by default" instruction:

- Load editor for bodyweight/mass/%max/%1RM/band/descriptive/unloaded.
- Effort selector adds RIR alongside RPE.
- Block-scoped authored-choice editor (add/remove choice, trigger text, applies-at step, options, and per-option actions for all seven `SessionChoiceAction` kinds).
- Per-step alternatives editor so `select_alternative` actions have a real target.

New pure factories in `sessions/sessionDraft.ts` (`createDraftChoice`/`Option`/`Alternative`/`Action`) carry default-value logic and are unit tested.

Deliberately deferred (recorded, not silently dropped): rest ranges, quality/technical-stop fields, and session-level `sessionTargets`/`prohibitedAdditions` remain JSON-import-only — still fully buildable via M3.6 JSON import, not a missing capability.

### DIFF
--- a/app/src/components/session/ManualSessionBuilder.tsx
+++ b/app/src/components/session/ManualSessionBuilder.tsx
@@ -174,12 +174,27 @@ export const ManualSessionBuilder: React.FC<ManualSessionBuilderProps> = ({ user
         }),
     }));
 
-    const removeAlternative = (blockIndex: number, stepIndex: number, altId: string) => setBlocks(current => current.map((block, index) => index !== blockIndex ? block : {
-        ...block,
-        steps: block.steps.map((step, stepAt) => stepAt !== stepIndex ? step : {
-            ...step,
-            alternatives: (step.alternatives ?? []).filter(alt => alt.id !== altId),
-        }),
+    // Also clears alternativeId on any select_alternative action targeting this step that
+    // referenced the removed alternative -- otherwise it dangles silently: validation only
+    // checks targetStepId/targetBlockId, and the runner's applySelectAlternative no-ops on an
+    // unresolvable id, so the athlete would pick "switch exercise" and nothing would happen.
+    const removeAlternative = (blockIndex: number, stepIndex: number, altId: string) => setBlocks(current => current.map((block, index) => {
+        if (index !== blockIndex) return block;
+        const removedFromStepId = block.steps[stepIndex]?.id;
+        return {
+            ...block,
+            steps: block.steps.map((step, stepAt) => stepAt !== stepIndex ? step : {
+                ...step,
+                alternatives: (step.alternatives ?? []).filter(alt => alt.id !== altId),
+            }),
+            optionSets: (block.optionSets ?? []).map(choice => choice.appliesAtStepId !== removedFromStepId ? choice : {
+                ...choice,
+                options: choice.options.map(option => ({
+                    ...option,
+                    actions: option.actions.map(action => action.kind === 'select_alternative' && action.alternativeId === altId ? { ...action, alternativeId: '' } : action),
+                })),
+            }),
+        };
     }));
 
     const updateAlternative = (blockIndex: number, stepIndex: number, altId: string, exerciseId: string | undefined, title: string) => setBlocks(current => current.map((block, index) => index !== blockIndex ? block : {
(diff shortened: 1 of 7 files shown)


app/src/components/session/ManualSessionBuilder.tsx
```diff
@@ -115,6 +160,107 @@ export const ManualSessionBuilder: React.FC<ManualSessionBuilderProps> = ({ user
...
+    // Authored choices (D-MCHOICE) live on the block, not the step, so a single choice can be
+    // remapped for multiple steps' alternatives -- but every fixture and this builder only ever
+    // scopes one to `appliesAtStepId`. Mutations are keyed by choice/option id, not index, since
+    // an option's action list can change length independently of its siblings.
+    const updateChoices = (blockIndex: number, updater: (choices: SessionChoice[]) => SessionChoice[]) => setBlocks(current => current.map((block, index) => index !== blockIndex ? block : {
+        ...block, optionSets: updater(block.optionSets ?? []),
+    }));
+
+    const addChoice = (blockIndex: number, appliesAtStepId: string) => updateChoices(blockIndex, choices => [...choices, createDraftChoice(appliesAtStepId, newId)]);
+    const removeChoice = (blockIndex: number, choiceId: string) => updateChoices(blockIndex, choices => choices.filter(choice => choice.id !== choiceId));
```
Szczepanov, line 219: `updateAction` merges a full replacement action into the old one instead of replacing it, so switching an action's kind via the dropdown at line 366 leaves stale fields from the previous kind on the saved object.

Example: an author adds an action defaulting to `reduce_load_percent` (`targetStepId: 'squat', percent: 10`), then switches its kind to `end_session` via the `<select>`. `createDraftAction('end_session', ...)` returns `{kind:'end_session'}`, but `updateAction` does `{...action, ...patch}`, producing `{kind:'end_session', targetStepId:'squat', percent:10}` — a spurious `targetStepId` survives on a kind that should carry none. If the author later removes step `squat`, `validateSessionDefinition` sees the dangling `targetStepId` and blocks Review with no UI field to fix it (end_session actions render no targetStepId control).

**Fixed**: added a `replaceAction` helper used specifically for the kind-change dropdown, which swaps the action outright; `updateAction` keeps its merge semantics for same-kind field patches (%, sets, reps, alternativeId).

---

app/src/components/session/ManualSessionBuilder.tsx
```diff
@@ -199,21 +347,84 @@ export const ManualSessionBuilder: React.FC<ManualSessionBuilderProps> = ({ user
...
-                                            <label>RPE<input type="number" min={1} max={10} step={0.5} placeholder="Optional" value={rpe} onChange={event => updateStep(blockIndex, stepIndex, { effort: event.target.value === '' ? undefined : { kind: 'rpe', target: Number(event.target.value) } })} /></label>
+                                            <label>Effort<select value={effortKind} onChange={event => changeEffort(blockIndex, stepIndex, event.target.value as 'none' | 'rpe' | 'rir')}><option value="none">Not tracked</option><option value="rpe">RPE</option><option value="rir">RIR</option></select></label>
+                                            {effortKind !== 'none' && <label>{effortKind === 'rpe' ? 'RPE target' : 'RIR target'}<input type="number" min={0} max={10} step={0.5} value={effortTarget} onChange={event => updateStep(blockIndex, stepIndex, { effort: { kind: effortKind, target: Number(event.target.value) || 0 } })} /></label>}
```
Szczepanov, line 351: Changing a choice's "Applies at" step doesn't update `targetStepId` on that choice's existing actions, so an option can silently act on a different step than the one the trigger question is now shown at.

Concretely: an option's `reduce_load_percent` action targets Step A (set from `appliesAtStepId` at creation time). The author changes "Applies at" to Step B — `updateChoice` only patches `choice.appliesAtStepId`, never the action's `targetStepId`. `validateSessionDefinition` only checks `targetStepId` is *some* valid step id in the whole definition, not that it matches `appliesAtStepId`, so this passes validation silently. At runtime the athlete is asked the trigger question in the context of Step B, but the action still reduces load on Step A — a wrong-movement mutation with no error or warning anywhere.

**Fixed**: added `updateChoiceAppliesAtStep`, which retargets every action across the choice's options that pointed at the previous step, and clears `select_alternative`'s `alternativeId` (alternatives are step-scoped, so an id valid on the old step is meaningless on the new one).

---

app/src/components/session/ManualSessionBuilder.tsx
```diff
@@ -115,6 +160,107 @@ export const ManualSessionBuilder: React.FC<ManualSessionBuilderProps> = ({ user
...
+
+    const addAlternative = (blockIndex: number, stepIndex: number) => setBlocks(current => current.map((block, index) => index !== blockIndex ? block : {
+        ...block,
+        steps: block.steps.map((step, stepAt) => stepAt !== stepIndex ? step : {
+            ...step,
+            alternatives: [...(step.alternatives ?? []), createDraftAlternative(undefined, 'Alternative movement', newId)],
+        }),
+    }));
+
+    // Also clears alternativeId on any select_alternative action targeting this step that
```
Szczepanov, line 177: `removeAlternative` deletes a step alternative without clearing any `select_alternative` action's `alternativeId` that referenced it, and `validation.ts` never checked `alternativeId` at all, so the dangling reference silently no-ops at execution.

An author adds an alternative, wires a `select_alternative` action to it, then later removes the alternative. `validateSessionDefinition` reports no error and the session saves. At execution, `choiceResolution.ts`'s `applySelectAlternative` does `step.alternatives?.find(c => c.id === alternativeId)`, finds nothing, and silently no-ops — the athlete picks "switch exercise" expecting a substitution and the app keeps the original (possibly contraindicated) movement with no error shown.

**Fixed**: `removeAlternative` now clears `alternativeId` on any action that referenced the removed alternative, and `validation.ts` now validates that a `select_alternative` action's `alternativeId` actually resolves to an alternative on its `targetStepId` (so a cleared/empty one blocks Review with a clear message instead of silently no-opping later).

---

app/src/components/session/ManualSessionBuilder.tsx
```diff
@@ -199,21 +347,84 @@ export const ManualSessionBuilder: React.FC<ManualSessionBuilderProps> = ({ user
...
                                             <label>Laterality<select value={step.laterality ?? 'bilateral'} onChange={event => updateStep(blockIndex, stepIndex, { laterality: event.target.value as SessionStep['laterality'] })}><option value="bilateral">Bilateral</option><option value="per_side">Per side</option><option value="alternating">Alternating sides</option></select></label>
                                             <label>Tempo<input value={step.tempo ?? ''} placeholder="e.g. 2-0-X-0" onChange={event => updateStep(blockIndex, stepIndex, { tempo: event.target.value || undefined })} /></label>
+                                            <label>Load<select value={loadKind} onChange={event => changeLoad(blockIndex, stepIndex, event.target.value as LoadKind)}>
+                                                {LOAD_KINDS.map(kind => <option key={kind} value={kind}>{LOAD_KIND_LABELS[kind]}</option>)}
+                                            </select></label>
```
Szczepanov, line 361: `removeOption` has no guard against deleting a choice's last remaining option, unlike `createDraftChoice` which guarantees at least one; `validation.ts` never rejected an empty `options` array either, so an author could save a trigger question with nothing to answer.

**Fixed**: the Remove button on an option is now disabled when it's the choice's only option, and `validation.ts` rejects a choice whose `options` array is empty.

---

Szczepanov (commented):
Reviewed the manual builder's new choice/action/alternative editor. Found four correctness gaps around the new UI letting authors reach states validation didn't check for — all fixed in commit 643992c, pushed to this branch (which also carries #96's merged-in review fixes):

1. Action-kind swap merged a patch instead of replacing, leaving stale fields from the previous kind.
2. Changing a choice's "Applies at" step left its actions' `targetStepId` pointing at the old step.
3. Removing a step alternative left a dangling `alternativeId` on any `select_alternative` action referencing it.
4. A choice's last option could be deleted, leaving a trigger question with nothing to answer.

(3) and (4) also needed matching `validation.ts` checks, since the underlying gaps were pre-existing there — just previously unreachable outside hand-written JSON.

---

PR #98, opened 2026-08-19
feat(sessions): companion occurrence start + Garmin dedup plumbing (M4.3)

Szczepanov:
Stacked on #97 (M3.8).

- `SessionRunner` offers a "Companion session available" prompt right after the primary session completes or is abandoned (never concurrently, per the single-active-execution runner model). Starting a companion resolves its `definitionRef` (known fixture or the athlete's own saved definition) and calls `startFixtureSession`/`startSession`, creating its own independent, no-authority execution; skipping records nothing. Embedded segments (`SessionDefinition.blocks`) were already structurally distinct from `companionSessions[]` and needed no change.
- New `sessions/occurrenceReconciliation.ts`: `sessionExecutionOccurrenceKey` gives every execution a stable identity; `matchExecutionsToGarminActivities` reconciles executions against Garmin activities by date/modality/duration tolerance so one physical occurrence isn't double-counted. Deliberately **not** wired into `buildTrainingHistorySnapshot`'s live engine path per D-MPOLICY (domain exposure from general session executions stays a default-off evidence candidate until its own ship decision) — forward-compatible plumbing, not activation.

### DIFF
--- a/app/src/components/session/SessionRunner.tsx
+++ b/app/src/components/session/SessionRunner.tsx
@@ -84,6 +84,16 @@ export const SessionRunner: React.FC<SessionRunnerProps> = ({
     const [savedDefinitions, setSavedDefinitions] = useState<SessionDefinitionHeader[]>([]);
     const [savedDefinitionsError, setSavedDefinitionsError] = useState<string | null>(null);
     const [startingSavedDefinitionId, setStartingSavedDefinitionId] = useState<string | null>(null);
+    // M4.3: a companion is a separately executable session referenced from the one that just
+    // finished (SessionDefinition.companionSessions), never an embedded block -- those already
+    // render inline within the same execution. Starting one creates its own execution; it may
+    // just as well be skipped. `finishedTitle` is only for the prompt's header copy.
+    const [companionPrompt, setCompanionPrompt] = useState<{
+        finishedTitle: string;
+        companions: NonNullable<SessionDefinition['companionSessions']>;
+    } | null>(null);
+    const [startingCompanionId, setStartingCompanionId] = useState<string | null>(null);
+    const [companionError, setCompanionError] = useState<string | null>(null);
     const [pendingGroupAdvance, setPendingGroupAdvance] = useState<{
         blockIndex: number;
         stepIndex: number;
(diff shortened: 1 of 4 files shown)


app/src/sessions/occurrenceReconciliation.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Finds the closest same-date, compatible-modality, comparable-duration Garmin activity for
+ * each execution -- one activity reconciles to at most one execution, mirroring
+ * `completedTraining.ts`'s own adherence-to-Garmin matching tolerance and exclusivity so the
+ * two reconciliation paths behave consistently. An execution with no resolved modality still
+ * matches on date and duration alone; an execution with no completed duration is never
+ * matched (nothing to compare yet, and "counted once" isn't at risk until it's counted at
+ * all).
+ *
+ * Matching is global, not per-execution-in-isolation: every valid (execution, activity) pair
```
Szczepanov, line 105: `matchExecutionsToGarminActivities`'s docstring says it "finds the closest" match, but which execution wins a contested Garmin activity is actually decided by arbitrary `executionId` sort order, not by which execution's duration is actually closer.

Two same-date, same-modality executions both within tolerance of one activity: `exec-a` duration 20, `exec-b` duration 29, activity duration 30. Because executions are processed in `executionId.localeCompare` order and `exec-a` sorts first, `exec-a` (diff=10) claims the activity on its first (and only, at that point) candidate — leaving `exec-b` (diff=1, the objectively closer match) unmatched. The module's own test claiming to verify "preferring the closer duration match" didn't actually catch this, since its example's alphabetically-first execution happened to also be the closer one.

**Fixed**: matching is now global — every valid `(execution, activity)` pair is a candidate, claimed in order of increasing duration difference, so a contested activity goes to the genuinely closer execution regardless of id order. Also corrected the test's fixture so it actually exercises this.

---

docs/plans/multidomain-session-authoring-execution-and-evidence.md
```diff
@@ -1240,6 +1240,50 @@ adapters, new `sessions/occurrenceReconciliation.ts`; UI companion card.
 **Done when.** An embedded bike warm-up stays inside Strength; a later recovery spin may be
 started or skipped independently; a matching Garmin ride is counted exactly once.
 
+**Outcome (2026-08-19).** The embedded-vs-companion distinction needed no new code: the model
+already keeps them apart structurally (`SessionDefinition.blocks` vs. `companionSessions[]`,
+rendered separately since M3.7), so fixture 01's embedded Olympic power block already stays
+inside its own execution with no change here.
+
+* **Starting a companion creates its own execution.** `SessionRunner.tsx` captures the
+  finishing session's `title`/`companionSessions` before `completeSession`/`abandonSession`
```
Szczepanov, line 1269: The M4.3 outcome note claims `completeSession`/`abandonSession` "clear `runner.definition`", but neither function calls `setRawDefinition` — verified in `useSessionRunner.ts`, where it's only called on session start/restore (and on a start failure), never inside `completeSession` or `abandonSession`. A future maintainer relying on "definition becomes null after completion" as an invariant would be wrong: the previous session's full definition remains in `runner.definition` until the next `startSession` overwrites it.

**Fixed**: corrected the note — the pre-completion capture of `title`/`companionSessions` in `SessionRunner.tsx` is what the companion prompt actually relies on, not any clearing behavior.

---

app/src/components/session/SessionRunner.tsx
```diff
@@ -179,6 +210,45 @@ export const SessionRunner: React.FC<SessionRunnerProps> = ({
         }
     };
 
+    /** Resolves a companion's `definitionRef` the same two ways the fixture/saved-session
+     * pickers above already resolve a startable definition -- a known reviewed fixture
+     * (covers the M0.2 corpus's recovery-spin companion) or one of the athlete's own saved
+     * manual definitions -- and starts it as its own independent, no-selection-authority
+     * execution (D-MAUTH `unplanned_log`), exactly like starting any other unplanned session. */
+    const startCompanion = async (companion: NonNullable<SessionDefinition['companionSessions']>[number]) => {
```
Szczepanov, line 218: The companion prompt let an athlete start a companion session immediately, ignoring `SessionDefinition.companionSessions[].notEarlierThanMinutesAfter` — a field that exists specifically to gate this.

A companion authored with `notEarlierThanMinutesAfter: 30` (e.g. a recovery spin that should follow a strength session by at least 30 minutes) could be started the instant the primary session finished — `startCompanion` and the "Start companion →" button never read the field, so the authored timing constraint was silently unenforced with no warning shown.

**Fixed**: the prompt now records `finishedAt`, disables "Start companion" (showing "Available in N minutes") until the gate opens, and `startCompanion` itself rejects an early start defensively.

---

Szczepanov (commented):
Reviewed the companion-prompt and Garmin-reconciliation code. Found two functional gaps plus a doc/code mismatch — all fixed in commit 7aa0463, pushed to this branch (which also carries #96/#97's merged-in review fixes):

1. `notEarlierThanMinutesAfter` was parsed onto the model but never enforced — a timing-gated companion could be started immediately.
2. `matchExecutionsToGarminActivities` decided contested-activity ownership by `executionId` sort order, not by actual duration closeness, despite its docstring claiming otherwise; the existing test didn't catch this because its fixture's sort-first execution happened to also be the closer one.
3. The M4.3 outcome note incorrectly claimed `completeSession`/`abandonSession` clear `runner.definition` — neither does.

Finding 2 is currently inert (this module isn't wired into any live engine path per the PR's own description) but worth having correct before M6.4/M8 activates it.

---

PR #99, opened 2026-08-19
feat(responses): occurrence-linked SessionResponse generalization (M5.1)

Szczepanov:
Stacked on #98 (M4.3).

New `responses/` domain, distinct from `sessions/` per D-MRECORDS:

- `responses/models.ts`: `SessionResponse`, `ResponseWindow` (`immediate|later_day|next_morning`), `SessionResponseSourceRef` — reuses M1.7/M2.6's existing `{kind, id, date}` linkage shape rather than migrating it, so no existing check-in document needs rewriting.
- `responses/validation.ts` + `persistence/parsers/sessionResponse.ts`: self-contained strict validator/parser mirroring `sessions/` conventions.
- `services/sessionResponseService.ts`: `recordResponse` (create-only, never fabricated for an unanswered prompt), `updateResponseFacts` (patches only the five non-tissue facts, preserving `sourceSession`/`window`/`date`/`createdAt`), `getResponsesForSource`/`getResponseForWindow` (`null`, not a default, when a window was never answered).
- `firestore.rules`: ownership, window/date/shape validation, an `occurrenceId` (when present) must resolve to a real occurrence under this same user's own path, and update rules reject any change to `sourceSession`/`window`/`date`/`createdAt` server-side.

Per D-MRESP, `DailySubjectiveCheckin.tissueResponses` remains the sole tissue authority — `SessionResponse` never stores a tissue value.

Extended `sessions/architecture.test.ts`'s optimizer/planner import boundary to also scan `responses/`, the second distinct-lifecycle domain now alongside `sessions/`.

82/82 Firestore emulator tests pass (5 new); full `npm run check` passes.

### DIFF
--- a/app/firestore.rules
+++ b/app/firestore.rules
@@ -908,6 +908,10 @@ service cloud.firestore {
         && isValidActivityDate(data.sourceSession.date)
         && data.window in ['immediate', 'later_day', 'next_morning']
         && isValidActivityDate(data.date)
+        // `date` equals sourceSession.date for `immediate`, strictly later otherwise
+        // (models.ts's own documented invariant) -- ISO YYYY-MM-DD strings compare
+        // lexicographically the same as chronologically.
+        && (data.window == 'immediate' ? data.date == data.sourceSession.date : data.date > data.sourceSession.date)
         && data.checkinRef is map
         && data.checkinRef.keys().hasOnly(['date'])
         && isValidActivityDate(data.checkinRef.date)
@@ -930,9 +934,13 @@ service cloud.firestore {
       allow create: if hasValidSessionResponse(userId, responseId);
       allow update: if hasValidSessionResponse(userId, responseId)
         && keepsOwnership(userId)
-        // Edits preserve provenance: identity/window/createdAt never change after
-        // creation -- only the non-tissue facts and updatedAt may be revised.
+        // Edits preserve provenance: identity/occurrence/window/createdAt never change
+        // after creation -- only the non-tissue facts and updatedAt may be revised.
+        // occurrenceId's presence and value are both pinned (not just its value when
+        // present), so a response created without one can't have one added later either.
         && request.resource.data.sourceSession == resource.data.sourceSession
+        && ('occurrenceId' in request.resource.data) == ('occurrenceId' in resource.data)
+        && (!('occurrenceId' in resource.data) || request.resource.data.occurrenceId == resource.data.occurrenceId)
         && request.resource.data.window == resource.data.window
         && request.resource.data.date == resource.data.date
         && request.resource.data.createdAt == resource.data.createdAt;
--- a/app/firestore.rules
+++ b/app/firestore.rules
@@ -891,6 +891,54 @@ service cloud.firestore {
       }
     }
 
+    // M5.1: SessionResponse -- linkage plus non-tissue session facts only (D-MRESP).
+    // DailySubjectiveCheckin.tissueResponses remains the sole tissue authority; this
+    // record never carries a tissue value.
+    function hasValidSessionResponse(userId, responseId) {
+      let data = request.resource.data;
+      return hasOwnedUserId(userId)
+        && data.keys().hasOnly(['userId', 'responseId', 'sourceSession', 'occurrenceId', 'window', 'date', 'checkinRef', 'sessionRpe', 'completedFraction', 'unexpectedFatigue', 'techniqueNote', 'note', 'createdAt', 'updatedAt'])
+        && data.keys().hasAll(['userId', 'responseId', 'sourceSession', 'window', 'date', 'checkinRef', 'createdAt', 'updatedAt'])
+        && data.responseId == responseId
+        && data.sourceSession is map
+        && data.sourceSession.keys().hasOnly(['kind', 'id', 'date'])
+        && data.sourceSession.keys().hasAll(['kind', 'id', 'date'])
+        && data.sourceSession.kind in ['strength', 'execution']
+        && data.sourceSession.id is string && data.sourceSession.id.size() > 0
+        && isValidActivityDate(data.sourceSession.date)
+        && data.window in ['immediate', 'later_day', 'next_morning']
+        && isValidActivityDate(data.date)
+        && data.checkinRef is map
+        && data.checkinRef.keys().hasOnly(['date'])
+        && isValidActivityDate(data.checkinRef.date)
+        && (!('occurrenceId' in data) || (
+          data.occurrenceId is string && data.occurrenceId.size() > 0
+          // Cannot reference another user's occurrence: it must exist under this same
+          // user's own path -- a cross-user id structurally cannot resolve here.
+          && exists(/databases/$(database)/documents/users/$(userId)/session_occurrences/$(data.occurrenceId))
+        ))
+        && (!('sessionRpe' in data) || (data.sessionRpe is number && data.sessionRpe >= 0 && data.sessionRpe <= 10))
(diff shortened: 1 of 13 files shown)


app/firestore.rules
```diff
@@ -891,6 +891,62 @@ service cloud.firestore {
...
+        && (!('unexpectedFatigue' in data) || data.unexpectedFatigue is bool)
+        && (!('techniqueNote' in data) || (data.techniqueNote is string && data.techniqueNote.size() <= 2000))
+        && (!('note' in data) || (data.note is string && data.note.size() <= 2000))
+        && data.createdAt is string && data.updatedAt is string;
+    }
+
+    match /users/{userId}/session_responses/{responseId} {
+      allow read: if isOwner(userId);
+      allow create: if hasValidSessionResponse(userId, responseId);
+      allow update: if hasValidSessionResponse(userId, responseId)
```
Szczepanov, line 935: The `session_responses` update rule pinned `sourceSession`, `window`, `date` and `createdAt` but never pinned `occurrenceId`, so it could be added or swapped after creation despite the module's stated provenance-immutability invariant.

A response first created with no `occurrenceId` (companion/unplanned execution) could later have one added via an update that only satisfies `hasValidSessionResponse`'s shape/existence checks — the update rule's equality checks never compared `request.resource.data.occurrenceId` to `resource.data.occurrenceId`, so the write would succeed and the response's selection-authority provenance (D-MAUTH) would silently change after the fact.

**Fixed**: the update rule now also pins `occurrenceId`'s presence and value, and `responses/validation.ts` + a new Firestore emulator test cover it.

---

app/src/services/sessionResponseService.ts
```diff
@@ -0,0 +1,146 @@
...
+        return responses.find(response => response.window === window) ?? null;
+    }
+
+    /** Creates a new response for a `(sourceSession, window)` pair that has never been
+     * answered before. Callers should still check `getResponseForWindow` first and call
+     * `updateResponseFacts` instead when one already exists -- but the deterministic id
+     * backs that up: a second `recordResponse` for the same pair (e.g. a double-tap/retry
+     * race) throws rather than silently overwriting an existing answer's `createdAt`/facts
+     * or leaving two documents behind. */
+    async recordResponse(
```
Szczepanov, line 98: `recordResponse` had no protection against creating two documents for the same `(sourceSession, window)` pair — `responseId` was freshly randomized on every call and nothing enforced uniqueness per window.

A double-tap submit (or any retry) calling `recordResponse` twice for the same execution+window before either write is reflected in a subsequent read would leave two `SessionResponse` docs for the same window; `getResponsesForSource`/`getResponseForWindow` would then pick whichever one `Array.prototype.find()` happened to encounter first, silently discarding one of the athlete's two answers with no error surfaced anywhere.

**Fixed**: `responseId` is now derived deterministically from `(sourceSession, window)`, and `recordResponse` checks for an existing document first and throws instead of silently overwriting or duplicating.

---

app/firestore.rules
```diff
@@ -891,6 +891,62 @@ service cloud.firestore {
...
+        && data.keys().hasOnly(['userId', 'responseId', 'sourceSession', 'occurrenceId', 'window', 'date', 'checkinRef', 'sessionRpe', 'completedFraction', 'unexpectedFatigue', 'techniqueNote', 'note', 'createdAt', 'updatedAt'])
+        && data.keys().hasAll(['userId', 'responseId', 'sourceSession', 'window', 'date', 'checkinRef', 'createdAt', 'updatedAt'])
+        && data.responseId == responseId
+        && data.sourceSession is map
+        && data.sourceSession.keys().hasOnly(['kind', 'id', 'date'])
+        && data.sourceSession.keys().hasAll(['kind', 'id', 'date'])
+        && data.sourceSession.kind in ['strength', 'execution']
+        && data.sourceSession.id is string && data.sourceSession.id.size() > 0
+        && isValidActivityDate(data.sourceSession.date)
+        && data.window in ['immediate', 'later_day', 'next_morning']
```
Szczepanov, line 909: Neither the Firestore rules nor the TS validator enforced the documented relationship between `window` and `date` (`date` must equal `sourceSession.date` for the `immediate` window, per `models.ts`'s own doc comment on `date`), so a logically inconsistent response would pass validation — e.g. `window: 'immediate'` with `date` several days after `sourceSession.date`. Any future M5.2/M5.3 consumer that assumed "immediate ⇒ date == sourceSession.date" would get silently wrong groupings.

**Fixed**: both `firestore.rules` and `responses/validation.ts` now enforce `date == sourceSession.date` for `immediate` and `date > sourceSession.date` otherwise.

---

app/src/responses/validation.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+    if (raw.completedFraction !== undefined && (typeof raw.completedFraction !== 'number' || raw.completedFraction < 0 || raw.completedFraction > 1)) {
+        issues.push({ path: 'completedFraction', message: 'completedFraction must be a number between 0 and 1' });
+    }
+    if (raw.unexpectedFatigue !== undefined && typeof raw.unexpectedFatigue !== 'boolean') {
+        issues.push({ path: 'unexpectedFatigue', message: 'unexpectedFatigue must be boolean' });
+    }
+    // 2000-char cap mirrors firestore.rules' size() <= 2000 check, so a too-long note fails
+    // here rather than surfacing as an unexplained permission-denied at write time.
+    if (raw.techniqueNote !== undefined && (typeof raw.techniqueNote !== 'string' || raw.techniqueNote.length > 2000)) {
```
Szczepanov, line 87: `validateSessionResponse` didn't enforce the 2000-character max on `techniqueNote`/`note` that `firestore.rules` requires, so the TS validator and the server-side rule disagreed on what's valid — a too-long note would pass client-side validation and only fail as an unexplained permission-denied at write time.

**Fixed**: added the matching 2000-char cap here.

Also reviewed: `matchExecutionsToGarminActivities`-adjacent `getResponsesForSource`'s window sort relied on `ResponseWindow` values happening to sort correctly alphabetically (`immediate < later_day < next_morning`) via `localeCompare` — undocumented and unenforced by any type-level mechanism. Replaced with an explicit order map so a future added/renamed window can't silently break chronological ordering.

Not changed: `isValidCalendarDate` duplicates `sessions/validation.ts`'s date-validation logic almost verbatim, but the module's own header comment explains this is deliberate ("Self-contained like `sessions/validation.ts` rather than importing its helpers -- `responses/` is its own distinct-lifecycle domain (D-MRECORDS)"), so I left it as-is rather than overriding that documented architectural decision.

---

Szczepanov (commented):
Reviewed the new `session_responses` Firestore rule, the `responses/` domain (models/validation), the persistence parser, and `SessionResponseService`. Found a real security/data-integrity gap plus a duplicate-document race, both fixed and pushed to this branch:

1. `occurrenceId` wasn't pinned on update, so a response's selection-authority provenance could silently change after creation despite the module's own stated invariant — the actual server-side enforcement gap, not just client discipline.
2. `recordResponse` had no protection against a double-tap/retry producing two documents for the same `(sourceSession, window)` pair, one of which any later read would silently discard.
3. Neither layer enforced the documented `window`/`date` relationship.
4. The TS validator's `techniqueNote`/`note` length check didn't match the Firestore rule's cap.

This branch also carries #96/#97/#98's review fixes merged forward, plus your own CodeRabbit-findings fix on #96 (merged in from the updated base).

`npm run check` and the Firestore emulator suite (83/83) both pass with the fixes applied.

---

PR #114, opened 2026-08-19
⚡ Bolt: [Performance Improvement] Batched Firestore Writes for Garmin Activities

Szczepanov:
💡 **What:** The optimization implemented a batched write operation for archiving Garmin activities. A new method `upsert_activities` handles batched chunking using Firestore's `db.batch()` API (up to 500 documents per batch) instead of processing individual document writes inside a loop.

🎯 **Why:** The previous code iterated over activities and called `upsert_activity` for each one, causing an N+1 network call problem when synchronizing many activities simultaneously. This slowed down the synchronization process.

📊 **Measured Improvement:** Measured with a 50-activity benchmark simulating a 50ms latency:
- **Baseline:** ~2.51s (50 separate network calls)
- **Improvement:** ~0.05s (1 batched network call)
- **Change over baseline:** ~50x speedup in the archiving method logic.

---
*PR created automatically by Jules for task [2616687135753290467](https://jules.google.com/task/2616687135753290467) started by @Szczepanov*


Szczepanov:
Reviewed and merged. The batched-write optimization is sound and the numbers check out. Only had a trivial conflict in `.jules/bolt.md` (two unrelated changelog entries added in parallel) against current `main` — merged `main` in, combined both entries, verified `ruff check`/`ruff format`/`mypy`/`pytest` all pass, then merged.

---

PR #115, opened 2026-08-19
🔒 Fix TOCTOU symlink vulnerability in token store permissions

Szczepanov:
🎯 **What:** Fixed a Time-of-Check to Time-of-Use (TOCTOU) symlink vulnerability in `src/garmin_sync/token_store.py` where `os.chmod` was used unsafely on potentially user-controlled paths.
⚠️ **Risk:** An attacker could replace the token directory or file with a symlink to another sensitive file/directory (e.g., `/etc/shadow`) just before `os.chmod` is called. This would allow them to change the permissions of the targeted file, potentially leading to unauthorized access or privilege escalation.
🛡️ **Solution:** Replaced `os.chmod` with a secure helper `_chmod_secure`. This function prioritizes using `os.chmod` with `follow_symlinks=False` if natively supported. If not, it uses `os.fchmod` combined with `os.open` and `os.O_NOFOLLOW` to prevent symlink traversal. A final fallback checks `path.is_symlink()` directly before applying permissions for older/unsupported platforms.

---
*PR created automatically by Jules for task [13672251761788548092](https://jules.google.com/task/13672251761788548092) started by @Szczepanov*


Szczepanov:
Reviewed and merged. Solid TOCTOU fix — `_chmod_secure` correctly layers `follow_symlinks=False` → `O_NOFOLLOW` fchmod → symlink-refusal fallback, and the exception log no longer leaks the raw error message. Merged cleanly against current `main` with no conflicts; verified `ruff`, `mypy`, and `pytest` all pass before merging.

---

PR #136, opened 2026-08-20
Robust biometric baseline observability + respiration comparison path

Szczepanov:
## Summary

This PR now separates **baseline observability** from **production decision policy**.

### Respiration — robust baseline candidate, production scoring default-off

- `src/garmin_sync/metrics.py` changes respiration's persisted v3 baseline from mean to **median**, with a 28-day **MAD** (`1.4826 × raw MAD`).
- The documentation now describes scaled MAD precisely: `1.4826` is a **normal-consistency factor**, not a guarantee that MAD is equivalent to standard deviation for bounded/skewed/quantized wearable data.
- `app/src/engine/rules.ts` retains a respiration strain implementation (`weight = 0.3`, adverse direction = elevated respiration) so the real scoring path can be exercised during replay/sensitivity work.
- `mapSnapshotToEngineInput()` exposes an explicit `RespirationStrainPolicy = 'off' | 'median-mad-v1'` and defaults to **`'off'`**. Existing production callers therefore do not make respiration decision-relevant yet.
- Even when the comparison policy is explicitly enabled, respiration is forwarded only when:
  1. `baselineComputationVersion >= 3`, and
  2. a real `respiration28dMad` exists.
  Legacy mean-based deltas and v3+ snapshots with unavailable MAD remain inert.
- The `1.0 br/min` floor and `0.3` weight are documented as **reference heuristics pending calibration**, not as validated Garmin measurement resolution or production parameters.

### Sleep / RHR / HRV / steps — observation-only candidate statistics

`BASELINE_COMPUTATION_VERSION` 4 adds 7d/28d median, 28d MAD, and matching median-baseline deltas alongside the existing live mean/stdev fields. Nothing in `rules.ts` or `fatigue.ts` consumes these new fields.

Importantly, this PR no longer assumes median/MAD is the eventual replacement for every metric. ADR-0024 records metric-specific candidates:

- **RHR:** robust median for acute anomaly detection vs mean/EWMA for gradual drift.
- **HRV:** compare current mean/stdev and median/MAD against a log-domain rolling mean + CV/SD approach; do not presume median/MAD wins.
- **Sleep score:** treat as a bounded vendor composite; compare personal baselines with absolute score, short multi-night history/debt, and trend/EWMA.
- **Steps:** separate habitual-activity baseline/surge detection from actual mechanical load; include weekday-aware baseline candidates.

### Body Battery / stress / Training Readiness — observation-only

`BASELINE_COMPUTATION_VERSION` 5 adds median/MAD observation fields for these metrics, but they remain unconsumed.

ADR-0024 explicitly documents their upstream dependence on HRV/sleep/stress/load and prohibits simply adding independent positive strain weights for all of them. A future implementation must compare either a weak Garmin composite prior or the project's own constituent-signal model rather than double-counting both.

## Research / architecture decision

New `docs/adr/0024-biometric-baseline-estimator-policy.md` records the research findings and release contract. It rejects a universal estimator migration and requires recorded replay evidence before any candidate estimator/weight becomes decision-relevant.

The required comparison includes, where applicable:

- mean-vs-median difference;
- stdev/MAD relationship;
- zero-MAD and ties/quantization frequency;
- missingness/coverage, skew/outliers, autocorrelation and weekday effects;
- cross-metric correlations;
- isolated and combined adverse-signal scenarios;
- recommendation mode flips and hard-session suppression;
- agreement with subjective readiness / explicit illness labels when available;
- lead time and false-positive persistence around known illness episodes.

Synthetic sweeps are useful for threshold discovery but are explicitly **not sufficient calibration** without historical replay or prospectively labelled data.

## Regression fixes added during review

- Pre-v3 respiration deltas were already present but mean-based. They are now version-gated rather than accidentally normalized with the MAD floor.
- Missing v3 MAD now means **zero respiration contribution**, rather than substituting a floor for an unavailable spread estimate.
- Adapter tests cover:
  - complete v3 data is inert under the production-default `off` policy;
  - pre-v3 data stays inert even when comparison scoring is requested;
  - v3 data without MAD stays inert;
  - v3 median/MAD values are forwarded only with explicit `median-mad-v1` opt-in.

## Documentation

- ADR-0006 was updated to describe respiration as a potentially useful **early anomaly/illness signal**, not a disease-specific detector.
- ADR-0006 now points to ADR-0024 for metric-specific estimator policy and calibration requirements.
- `metrics.py` comments/docstrings now avoid treating scaled MAD as a universal drop-in standard deviation replacement.

## Validation

The branch had green Python/TypeScript/unit/simulation validation before the review amendments. The latest CI run should be treated as the authoritative validation for the amended head; do not infer safety from an unchanged simulation corpus alone.

## Domain invariants

- [x] Firestore writes remain user-scoped under `users/{APP_USER_ID}/...`.
- [x] Calendar-date semantics remain `Europe/Warsaw`; completed `totalSteps` still means `D - 1`.
- [x] New sleep/RHR/HRV/steps/body-battery/stress/readiness baseline fields remain observation-only.
- [x] Respiration production consumption is default-off until the ADR-0024 replay/calibration gate is satisfied.


Szczepanov:
Review notes — overall direction looks good, especially the robust respiration baseline and the observation-only posture for the other metrics. CI is green. I’d address two things before merge:

1. **[Blocking] Pre-v3 compatibility is currently not inert as documented.**

   `respirationVs7d` / `respirationVs28d` already exist on pre-v3 snapshots; only `respiration28dMad` is new in v3. `mapSnapshotToEngineInput()` forwards the existing deltas unconditionally, while `metricStrain()` replaces a missing MAD with `RESPIRATION_MAD_FLOOR_BR = 1.0`.

   That means an old, mean-based snapshot can suddenly become decision-relevant. Example: `delta7 = +2`, `delta28 = +3`, MAD absent gives roughly `0.6` acute + `0.45` chronic = `1.05` respiration strain, enough by itself to cross the `modify` threshold of `1.0`.

   **Recommendation:** gate all three respiration engine fields on `baselineComputationVersion >= 3` (and probably MAD availability if the intended contract is “missing signal => zero contribution”), e.g. at the adapter/version boundary. Alternatively, make respiration strain explicitly return zero unless the v3 spread statistic is present. I’d also add an adapter/evaluator integration test using a realistic v2 snapshot with non-null respiration deltas but no MAD. The current legacy test deletes `respiration_delta`, which is not the actual pre-v3 document shape and therefore misses this case.

2. **[Calibration/release gate] The live `0.3` respiration weight needs evidence that actually exercises the new path.**

   The PR correctly calls the weight a first-pass heuristic and says it has not gone through the sensitivity/simulation harness, while `simulate:diff` reports “No semantic differences found.” For a decision-affecting change, that strongly suggests the current scenario corpus is not exercising respiration rather than proving the change is neutral.

   I’d apply the same evidence bar used elsewhere for fusion-policy changes: either keep respiration consumption behind a default-off selector until historical replay/sensitivity is recorded, or extend the harness with realistic `delta7` / `delta28` / MAD ranges and record how often modes change before enabling it. Please include combined cases where respiration, RHR, HRV and sleep all move adversely — those signals can be correlated, and the additive sum is where threshold crossings/double-counting risk matters most.

**Nice-to-have:** eventually consider explicit `respiration7dMedian` / `respiration28dMedian` field names instead of changing the semantics of `*Avg` by version. The current versioning/docs make it workable, but it is easy for a future reader to consume the field without checking the version.

Everything else looks structurally disciplined: v4/v5 additions remain observation-only, they are not accidentally threaded into `EngineObjectiveInput`, and the policy-version bump is present.

Szczepanov:
@coderabbitai review

Szczepanov:
## Research follow-up: estimator choice should be metric-specific

I did a deeper methods/literature review of **mean/SD vs median/MAD** for RHR, HRV, sleep, respiration, steps, and the newer Garmin-derived metrics in this PR. My conclusion is a bit different from the generalization implied by the PR text:

> **Median/MAD is a useful additional candidate estimator, but there is no good evidence for one universal baseline statistic across these metrics.**

The additive/observation-only posture for v4/v5 is therefore the right architecture. I would keep those fields, but I would make the documentation more precise and build the future comparison harness around **metric-specific candidates**, not “mean/SD vs median/MAD, winner takes all.”

### 1. Statistical point: `1.4826 * MAD` is only SD-consistent under a Gaussian model

Median/MAD has excellent contamination resistance. That is exactly why it is attractive when a trailing baseline can contain the abnormal episode we are trying to detect. But the current docstring wording that scaled MAD is generally “stdev-comparable” is too broad.

NIST explicitly notes that SD is efficient under normality while MAD/IQR are more robust for non-normal/long-tailed data. The `1.4826` multiplier makes MAD comparable to sigma **for Gaussian data**; it does not make MAD a universal SD-equivalent for bounded, skewed or quantized metrics. NIST also documents `Qn` as a robust scale alternative with the same 50% breakdown point, much higher Gaussian efficiency (~82% vs ~37% for MAD), and less dependence on symmetry.

Implications for this PR:

- rename/document scaled MAD as **“normal-consistent robust scale estimate”**, not universally “stdev-comparable”;
- do not assume a `MAD-z` has the same calibration across respiration, HRV, steps and 0–100 scores;
- measure the **zero-MAD rate**. Quantized/stable wearables can easily produce MAD = 0, in which case the configured floor—not the athlete’s observed variability—controls sensitivity;
- for the future comparison harness, consider `Qn`/trimmed estimators as a third candidate, not necessarily something to add to production now.

Sources: [NIST Measures of Scale](https://www.itl.nist.gov/div898/handbook/eda/section3/eda356.htm), [NIST Qn scale](https://itl.nist.gov/div898/software/dataplot/refman2/auxillar/qn_scale.htm), [NIST note on 1.4826 MAD being Gaussian-comparable](https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=913874).

---

## Metric-by-metric findings

### Respiration rate — **median/MAD is a defensible candidate, but not yet a validated winner**

This is the metric for which the robust-baseline argument is strongest. Wearable studies show nocturnal respiration is relatively stable within a person and often rises with respiratory illness. In a large Fitbit analysis, wearable RR had RMSE ~0.65 and MAE ~0.46 breaths/min against reference measurements; symptomatic COVID cases frequently showed meaningful RR elevation. In a small NCAA female-athlete cohort, RR was already elevated 3 days before a positive test, earlier than RHR/HRV in that sample.

However, two claims in the PR should be softened:

1. **“one of the earliest, most specific illness signals”** is too strong. RR is responsive to infection, but also to fever, heat, altitude, sleep-disordered breathing, autonomic stress, etc.; the Fitbit illness classifier could not distinguish COVID from influenza. Better wording: **“a relatively stable personal metric that can change early during respiratory/systemic illness.”**
2. I could not find evidence for **“Garmin’s ~1 br/min effective rounding resolution.”** Published wearable validation error around ~0.5–0.75 br/min can justify testing a 1.0 floor as a conservative heuristic, but that is not the same as device resolution.

Recommendation: keep the 28d median/MAD candidate, but treat `RESPIRATION_MAD_FLOOR_BR = 1.0` and weight `0.3` as calibration parameters. I would sensitivity-test floors `{0.5, 0.75, 1.0, 1.25, 1.5}` and several weights on real history before declaring this live policy calibrated.

Also note an independence issue: Garmin says HRV is used in respiration estimation, so HRV strain + respiration strain are **not guaranteed independent evidence**. That does not make RR useless; it means the harness should measure their correlation before summing fixed weights.

Sources: [Fitbit RR validation / COVID application](https://www.nature.com/articles/s41746-021-00493-6), [Fitbit physiological signs and illness](https://www.nature.com/articles/s41746-020-00363-7), [female-athlete COVID study](https://pubmed.ncbi.nlm.nih.gov/37401442/), [Garmin respiration technology](https://www.garmin.com/en-GB/garmin-technology/health-science/respiration-rate/), [Garmin HRV support – HRV contributes to respiration](https://support.garmin.com/en-GB/?faq=04pnPSBTYSAYL9FylZoUl5).

### Resting HR — **robust median has particularly good support for anomaly detection**

RHR is much more stable within a person than between people. A Nature Medicine real-time alerting study used the **median of average overnight RHR** as the personal baseline; >80% of participants had a stable baseline after seven nights. Other wearable infection work has also used a personal median baseline for RHR, sleep and activity.

So for **illness/anomaly detection**, long-window median is very defensible. For **training adaptation/trend**, rolling mean remains useful because gradual changes are signal, not contamination. Garmin itself exposes daily RHR plus a rolling 7-day average; Garmin daily RHR is already a special statistic—the lowest 30-minute average in 24 h.

Recommendation: keep both. A plausible future policy is **long-term robust baseline for anomaly detection + short-term mean/EWMA for drift**, rather than replacing all RHR mean/SD fields with median/MAD.

Sources: [NightSignal, Nature Medicine](https://www.nature.com/articles/s41591-021-01593-2), [wearable data and viral illness](https://www.nature.com/articles/s41591-020-1123-x), [Garmin RHR calculation](https://support.garmin.com/en-GB/?faq=F8YKCB4CJd5PG0DR9ICV3A).

### HRV — **the strongest counterexample to a universal median/MAD cutover**

Athlete-monitoring literature repeatedly favors **frequent measurements, log-transformed RMSSD (LnRMSSD), weekly averages and weekly CV**. A 2025 narrative review again highlights routine near-daily HRV, weekly averages and CV. Older athlete studies similarly use 7-day rolling means of LnRMSSD and its CV because both the level and the *variation in variability* can be informative.

Garmin’s own HRV Status also uses a **7-day average** against a slowly adapting personal baseline, requires about **three weeks** of sleep data to establish that baseline, and considers both unusually low **and unusually high** HRV potentially “unbalanced.”

Important caveat: we should first confirm exactly what Garmin `hrvSummary.lastNightAvg` represents statistically before directly importing LnRMSSD research assumptions. Garmin publicly calls it whole-night average HRV but does not make every implementation detail explicit.

Recommendation:

- **do not cut HRV from raw mean/SD to raw median/MAD merely for consistency with respiration**;
- add a future observation candidate for `log(hrvOvernightAvg)` + rolling mean/SD or CV *if* provenance confirms that transformation is meaningful for the exported metric;
- compare raw mean/SD, raw median/MAD, and log-domain weekly mean/CV;
- consider a future two-sided/contextual HRV anomaly model rather than permanently assuming “higher can never be adverse.”

Sources: [2025 HRV athlete-monitoring review](https://pubmed.ncbi.nlm.nih.gov/41516438/), [minimum measurement days / LnRMSSD averaging](https://pubmed.ncbi.nlm.nih.gov/24334285/), [elite triathlete 7-day LnRMSSD + CV](https://pubmed.ncbi.nlm.nih.gov/22367011/), [Garmin HRV baseline and 7-day average](https://www.garmin.com/en-GB/garmin-technology/health-science/hrv-status/), [Garmin HRV baseline support](https://support.garmin.com/en-GB/?faq=HnFAR4oFRF4kHeqYme3bU6).

### Sleep score — **estimator choice is secondary to measurement/composite semantics**

Garmin Sleep Score is a bounded 0–100 proprietary composite. Garmin says it incorporates duration, sleep architecture, HRV-derived stress/recovery, interruptions/restlessness and other factors. Consumer wearables are useful longitudinally, but meta-analyses and Garmin-specific PSG comparisons show non-trivial error in total sleep time, wake detection and especially sleep stages.

That means a Gaussian-like z-score interpretation of `1.4826*MAD` is particularly weak here. Median/MAD is fine as an **observational personal-baseline descriptor**, but I would not assume it is better than mean/SD for coaching decisions.

For actual decision logic, **last-night absolute score + multi-night sleep history/debt** may be more physiologically interpretable than a single robust z-score. Garmin’s own Training Readiness explicitly uses both last-night Sleep Score and the previous 3 nights of sleep history.

Sources: [Garmin Sleep Score calculation](https://www.garmin.com/en-IE/garmin-technology/health-science/sleep-tracking/), [2025 consumer wearable vs PSG meta-analysis](https://pubmed.ncbi.nlm.nih.gov/39484805/), [Garmin Forerunner 945 vs PSG](https://pubmed.ncbi.nlm.nih.gov/36502241/), [Garmin Training Readiness](https://www8.garmin.com/manuals/webhelp/GUID-2C274FD2-F0C3-445C-B0AC-700FECCE12E9/EN-US/GUID-C21BE0C8-A08E-4DA1-B6C6-2E0E2DDDB372.html).

### Steps — **habit baseline and load baseline are different problems**

Daily steps have substantial day-to-day behavioral variability. A 212k-person study found that multiple days are needed to reliably estimate habitual mean daily steps. Robust medians are also perfectly reasonable when the question is “what is this person’s typical activity day?”

But a 25–30k-step day is often **real mechanical load**, not an outlier to suppress. Therefore:

- for **habit/anomaly detection**: median/IQR/MAD can be useful;
- for **accumulated load/fatigue**: mean, sum or EWMA is conceptually preferable because large days should retain their dose;
- for this project’s current **ambient-step surge** rule, comparing both mean and median baselines is worthwhile, but I would also test a **day-of-week-aware** baseline because weekend hikes vs desk-work weekdays are structured behavior, not noise.

The current `fatigue.ts` rule is exactly a surge detector (`ambientSteps >= 1.8 × 7d baseline` plus `+6000`), so median may eventually improve the *reference habit level*, but that should be proven by replay; it should not imply using median for external load accumulation.

Source: [212,048-adult step reliability study](https://www.nature.com/articles/s41598-021-89141-3).

### Body Battery, Stress, Training Readiness — **descriptive baselines only until dependency/double-counting is handled**

This is the most important architectural finding for commit 3. These are **not independent sensor biomarkers**:

- Garmin Stress is primarily HR + HRV;
- Body Battery is driven by stress/HRV, sleep/rest and activity;
- Training Readiness explicitly fuses Sleep Score, Recovery Time, HRV Status, Acute Load, 3-night Sleep History and 3-day Stress History;
- Sleep Score itself already contains HRV-derived stress/recovery.

So a future engine that adds `HRV strain + sleep strain + stress strain + Body Battery strain + Training Readiness strain` would double/triple-count the same upstream physiology and training load.

Recommendation: the current **observation-only** status is correct. Document these fields as **correlated/composite evidence**. If Training Readiness is ever consumed, choose one of two architectures: either use Garmin’s composite as a weak “expert prior”, **or** use the constituent raw metrics; do not independently sum both without an explicit learned/calibrated fusion model.

For the baseline statistic itself, median/quantile bands are reasonable descriptors for bounded 0–100 scores. `stressMax` deserves extra caution because it is already an extreme statistic per day; taking a median across daily maxima is descriptive, but a MAD-z interpretation has no obvious physiological calibration.

Sources: [Garmin Training Readiness inputs](https://www8.garmin.com/manuals/webhelp/GUID-2C274FD2-F0C3-445C-B0AC-700FECCE12E9/EN-US/GUID-C21BE0C8-A08E-4DA1-B6C6-2E0E2DDDB372.html), [Garmin Training Readiness technology](https://www.garmin.com/pt-BR/garmin-technology/running-science/physiological-measurements/training-readiness/), [Garmin Stress](https://www.garmin.com/en-US/garmin-technology/health-science/stress-tracking/), [Garmin Body Battery](https://www.garmin.com/en-CA/garmin-technology/health-science/body-battery/), [Garmin HRV dependency list](https://support.garmin.com/en-GB/?faq=04pnPSBTYSAYL9FylZoUl5).

### Other metrics I would treat differently if/when added

- **Sleep duration:** use rolling mean/EWMA and cumulative sleep-debt logic; median can hide repeated moderate short-sleep nights.
- **Training load / activity dose:** sum/EWMA/decay, not median. A high-load day is the quantity we want to preserve.
- **Skin temperature:** robust personal deviation baseline is a good candidate because the use case is anomaly detection, but calibrate against the actual device’s measurement behavior.
- **SpO2:** bounded/ceiling-heavy and artifact-prone; lower-tail quantiles/episodes are more natural candidates than a symmetric mean±SD or MAD-z model.

---

## What I recommend changing/adding around this PR

### Keep

1. **Keep all v4/v5 median/MAD fields observation-only.** This is useful groundwork.
2. **Keep respiration median/MAD as a candidate**; the contamination argument is sound.

### Change before treating the research rationale as settled

3. Change `calculate_mad`/ADR wording from “stdev-comparable” to something like **“scaled by 1.4826 to be normal-consistent; robust scale, not assumed Gaussian in downstream interpretation.”**
4. Replace **“earliest, most specific illness signal”** with a more defensible RR claim.
5. Remove or qualify **“Garmin ~1 br/min effective rounding resolution”** unless we have a Garmin-specific source. Treat 1.0 as a policy floor chosen for calibration.
6. Add a warning to v5 docs/types that **Stress / Body Battery / Training Readiness / Sleep Score overlap strongly with HRV and each other** and must not later be summed as independent strains.
7. For HRV, make **log-domain rolling mean/CV** an explicit future candidate. Median/MAD should not become the default simply because it is robust.
8. For steps, explicitly distinguish **habit baseline** from **load aggregation** and include a weekday-aware candidate in replay.

### Comparison harness I would require before any v4/v5 cutover

For each metric, persist/report at least:

- valid-day count and missingness;
- mean–median gap;
- SD / scaled-MAD ratio;
- % windows with MAD = 0 / near zero;
- quantization/tie rate;
- skewness / upper- and lower-tail behavior;
- 7d/28d autocorrelation and trend;
- weekday effect for steps;
- cross-metric correlation, especially HRV↔RR↔stress↔sleep↔Body Battery↔Training Readiness.

Then replay candidate policies:

1. current mean + SD;
2. median + scaled MAD;
3. trimmed/winsorized mean + robust scale (`Qn` is worth benchmarking);
4. HRV-specific log-domain mean + CV/SD;
5. EWMA/dual-timescale candidates where gradual adaptation matters;
6. weekday-aware steps baseline.

And score **decision quality**, not just statistic agreement: mode flips, recovery-share changes, hard-session suppressions, false positives on clearly-green check-in days, agreement with `illnessSymptoms`/fatigue/readiness, lead time around known illness episodes, and sensitivity to floors/weights.

---

## One existing blocker remains

The earlier **pre-v3 respiration version-gating issue is still valid on the current head**: pre-v3 snapshots already contain mean-based respiration deltas, and a missing MAD currently falls back to the 1.0 floor, so legacy data can become decision-relevant. The unresolved inline review thread has the concrete fix/test recommendation.

More broadly, the current PR says `simulate:diff` finds **no semantic differences** while respiration is a live decision-affecting signal. I would treat that as evidence that the scenario corpus is not exercising the new signal, not as validation of weight `0.3`. Before release, either keep respiration wiring behind an experimental/default-off policy selector or extend the harness with realistic respiration trajectories and run a sensitivity/replay comparison.

### Bottom line

**I like the schema/observability direction. I would not remove the median/MAD additions. But I would explicitly reject a universal future median/MAD cutover and make the estimator policy metric-specific.** Respiration and RHR have the best robust-baseline case; HRV has strong evidence for log-transformed rolling averages/CV; sleep score and Garmin readiness composites need cautious bounded/composite treatment; steps need separate “habit” vs “dose” semantics.

Szczepanov:
Implemented the research-review recommendations on the PR branch.

Key changes:

- **Respiration scoring is now production-default `off`.** `mapSnapshotToEngineInput()` gained `RespirationStrainPolicy = 'off' | 'median-mad-v1'`; only explicit comparison/replay opt-in can forward respiration strain inputs.
- **Legacy/version safety fixed.** Pre-v3 mean-based respiration deltas are always inert. v3+ snapshots without a measured `respiration28dMad` are also inert, so the 1.0 floor cannot fabricate a spread estimate.
- **Regression coverage added** for default-off, pre-v3, missing-MAD, and explicit v3 median/MAD opt-in cases.
- Added **ADR-0024: Metric-specific biometric baseline estimators**. It rejects a universal median/MAD migration and records the candidate policies for respiration, RHR, HRV, sleep score, steps, Body Battery, stress, and Training Readiness, together with replay/release criteria and literature/provider references.
- Updated **ADR-0006** so respiration is described as a potentially useful early anomaly/illness signal rather than a disease-specific detector, and so the `0.3` weight / `1.0 br/min` floor are explicitly reference heuristics pending calibration.
- Updated `metrics.py` documentation: `1.4826 × MAD` is described as a **normal-consistent** robust scale estimator, not a universal SD-equivalent/drop-in denominator; zero MAD / ties are called out explicitly.
- The PR title/body have been rewritten to reflect the new architecture and no longer claim respiration is already live in production.

I deliberately did **not** implement HRV log-mean/CV, RHR EWMA, or weekday-aware steps as new production rules in this PR. ADR-0024 defines them as comparison candidates because selecting one before replay would repeat the same evidence problem we are fixing here.

**Validation on current head `42c2bff`: fully green.** Python pre-commit/lint/format/mypy/pytest/dependency audit, Docker build, frontend lint/tests/coverage/build/dependency audit, workout-catalog validation, policy-version drift check, Firestore emulator tests, simulation aggregate gate, and simulation semantic-diff all passed.

Szczepanov:
Follow-up observability update implemented on the PR branch:

- **Data View:** now shows raw stress avg/max and device Training Readiness, baseline computation version, version-correct respiration labels (pre-v3 values are explicitly marked legacy mean baselines), and a compact **Observation-only Candidate Baselines** block for sleep, RHR, HRV, steps, Body Battery wake, stress avg/max, and Training Readiness (`7d median · 28d median · MAD · Δ7 · Δ28`). It also states that respiration scoring is default-off/comparison-only and that the v4/v5 candidates are not extra engine strain inputs.
- **AI Context brief/export:** now exports the v3 respiration robust baseline and the full v4/v5 candidate set from the latest recovery snapshot. The text explicitly distinguishes live mean/stdev inputs from observation-only median/MAD candidates, states that production respiration scoring is OFF by default, and warns the planner not to add Body Battery/stress/Training Readiness as independent evidence because of upstream correlation/double-counting.
- **Regression coverage:** added `contextBrief.biometric-observability.test.ts`, including a v5 full-export case and a pre-v3 case proving legacy respiration means are not mislabelled as robust medians.

This is observability/export only; it does not opt any new candidate baseline into production decision scoring.

PR #141, opened 2026-08-20
Make context brief a safe collaborative planning handoff

Szczepanov:
## Why

The context brief is pasted into a fresh AI chat and used for day-by-day planning. On `main`, it is strong retrospectively but has three important handoff gaps:

1. source failures are shown in DataView UI but are **not copied into the brief text**, so an external planner can read an unavailable activity source as “No recorded sessions”;
2. it does not expose **already-planned future state** (fixed activities / imported plan sessions), so a fresh chat can unknowingly plan over a match or existing session;
3. the final `Requested output` block acts like an embedded command, which can make a fresh model generate a plan immediately instead of treating the pasted document as context for the user’s next question.

## Changes

- adds a planning-handoff layer around the existing tested retrospective brief rather than rewriting baseline/recovery logic;
- copied text now includes:
  - planning date and planning mode;
  - latest Garmin snapshot/sync timestamp and metric freshness dates;
  - explicit current-day check-in freshness;
  - same-day activity sync caveat;
  - current app recommendation, labelled as advisory rather than authority;
  - **DATA INCOMPLETE** warning inside the copied text whenever a source is unreadable;
  - a compact 7-calendar-day cross-signal timeline for sleep/HRV/RHR/respiration/Body Battery/stress/readiness/fatigue/soreness/training/flags;
  - next 7 days of fixed activities and active imported-plan sessions, including dose, priority/flexibility and moved/event status;
- future external-plan placement fails closed when fixed-activity occupancy cannot be read;
- unreadable active-goal state is now surfaced instead of silently becoming “no goals”;
- replaces the unconditional `Requested output` command with `How to use this handoff`:
  - answer the user’s actual question first;
  - current illness/pain/tissue response outranks favorable wearables;
  - do not add load solely because wearables look green;
  - preserve future key-session intent and hard/easy spacing;
  - preserve fixed/imported sessions unless the user asks to change them;
  - the exact importable day-block schema remains available conditionally when requested.

## Scope

The existing `buildContextBrief()` retrospective renderer, biometric estimator semantics and detailed Garmin activity telemetry remain unchanged. The new layer is intentionally additive/post-processing so the already-validated baseline logic stays stable.

## Tests

Adds focused unit coverage for:
- copied source-failure warnings;
- stale/current wearable and check-in semantics;
- 7-day cross-signal timeline;
- fixed and imported future sessions;
- externally-planned mode with no visible session;
- service future-range resolution and fail-closed occupancy behavior.

## Summary by CodeRabbit

* **New Features**
  * Added planning-focused context briefs with current data freshness, planning mode, sensor and preference status, goals, commitments, imported sessions, travel details, and recovery timelines.
  * Added clear notices when information is incomplete, unavailable, stale, or uncertain.
  * Improved handling of upcoming schedules and external plans, including safe fallbacks when details cannot be confirmed.
  * Preserved retrospective context while adding actionable planning guidance and usage instructions.
* **Bug Fixes**
  * Improved reporting for unreadable goals, overlays, check-ins, and incomplete data.

### DIFF
--- a/app/src/engine/contextBriefPlanningHandoff.test.ts
+++ b/app/src/engine/contextBriefPlanningHandoff.test.ts
@@ -218,6 +218,14 @@ function handoffInput(overrides: Partial<ContextBriefPlanningHandoffInput> = {})
             status: 'planned',
             moved: false,
             isEvent: false,
+            prescription: {
+                summary: '3 x 10 min threshold with controlled recovery',
+                steps: [
+                    { name: 'Warm-up', durationMin: 15, target: 'easy' },
+                    { name: 'Threshold', sets: 3, durationMin: 10, target: 'RPE 7–8', recoverySec: 240 },
+                    { name: 'Cool-down', durationMin: 10, target: 'easy' },
+                ],
+            },
         }],
         unavailableSources: [],
         ...overrides,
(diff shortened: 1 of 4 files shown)


Szczepanov (commented):
Reviewed this. The planning-handoff layer itself is solid and well-tested, but I found a few gaps and opened #145 (draft, targeting this PR's branch) with fixes + new tests:

1. **This branch currently fails `npm run check`.** `planningModeArchitecture.test.ts` (the ADR-0017 whole-app guard) flags `contextBriefPlanningHandoff.ts` for reading a field literally named `planningMode` — the guard matches by property name, not type, and can't tell that this field is already-resolved `PlanningContext.mode`, not a re-read of `TrainingIntentProfile.planningMode`. Fixed by renaming the field to `effectivePlanningMode`.
2. **Data loss in the copied prescription detail:** `renderPrescriptionStep()` handles `recoverySec`/`setRecoverySec` but drops `recoveryMin`/`setRecoveryMin` entirely, even though both are real, actively-used fields on `ExternalPrescriptionStep`. An imported interval session with minute-denominated recovery would show up in the handoff missing its recovery dose.
3. **A confirmed-absence claim that isn't actually confirmed:** when today's own `getActivePlanState` read fails (rejects, or returns a non-`AVAILABLE`/non-`MISSING` status) while fixed-activity occupancy is otherwise readable, `currentExternalSession` silently stays `null` and the brief renders "external-plan fallback today: no imported session is placed on this date" as fact — the same failure mode this PR explicitly closes for the occupancy-unreadable case, just one layer down. Now labelled `UNCONFIRMED` in that case.
4. Minor: `planningAuthoritySession()` reimplemented the v1/v2 discriminator inline instead of reusing the existing `isV2Session()` helper.

Details and tests are in #145.

---

PR #155, opened 2026-08-21
feat: persist performance outcome observations

Szczepanov:
## Scope

Implements PR B / `OV2.1–OV2.4` from `docs/plans/performance-outcome-validation.md`: the raw performance-outcome persistence and manual-entry foundation. This remains evidence-only and does not add progress labels, block verdicts, testing UI, or recommendation-policy behavior.

### Included
- OV2 evidence record contracts and fail-closed runtime validation.
- User-scoped immutable `MeasurementProtocol` revisions, completing the persistence side of OV1.2.
- Deterministic manual observation adapter: exact metric/unit/protocol/context validation plus comparison-series keying.
- Stable logical observation identity (`assessmentAttemptId:metricId`) with immutable revisions and a transactional mutable head.
- Explicit correction semantics: revision `N+1` supersedes `N`; stale/skipped head advances fail rather than overwrite evidence.
- Assessment-attempt lifecycle and purpose/protocol provenance.
- First-class immutable ecological `CompetitionOutcome` records kept outside protocol comparison series.
- Derived-observation provenance requiring source observation IDs plus algorithm version.
- Firestore rules for protocol/observation/attempt/competition collections.
- Dedicated emulator coverage for user isolation, malformed/unit-invalid writes, immutable protocol/revision denial, atomic initial writes, valid correction chains, stale correction denial, assessment provenance, and ecological-outcome isolation.

### Documentation / ownership reconciliation
- `docs/plans/performance-outcome-validation.md` is `In progress`; OV0.1–OV2.4 are complete on this branch.
- PR A / #154 is recorded as the merged OV0/OV1 contract slice; this PR is the OV2 persistence/manual-entry slice.
- The repeated-standardized-testing trigger originally recorded under M7 fired on 2026-08-21. **Implementation and live status ownership now belong exclusively to OV.**
- `docs/plans/multidomain-session-authoring-execution-and-evidence.md` no longer contains actionable M7.1–M7.4 task-board rows or M7.3/M7.4 implementation specs. It retains only a historical mapping from the former M7 requirements to their canonical OV owners.
- `docs/plans/README.md` no longer lists an OV/M7 item as M work. M's current dependency-unblocked item is M8.1; OV's dependency-unblocked items are OV3.1, OV5.1 and OV5.2. The next **planned** OV code slice remains OV3.1–OV3.4 / PR C.
[...]

PR #158, opened 2026-08-21
fix(observations): stabilize OV2 emulator suite and close rule gaps

Szczepanov:
Review/fix pass on #155.

## Summary

- Root-caused and fixed the failing "Frontend Build & Firestore Rules" CI job on #155: `performanceOutcomeRules.emulator.test.ts` shared a Firestore emulator project id with `firestoreRules.emulator.test.ts`. Vitest runs emulator test files in parallel by default, and both suites call `clearFirestore()` in their own `afterEach`, so one file's cleanup could wipe documents mid-transaction in the other — producing the intermittent `PERMISSION_DENIED` / "evaluation error" failures seen on CI. Confirmed by running the new suite standalone (passed) vs. alongside the other file (failed non-deterministically, including a different manifestation each run). Fix: give the new suite its own emulator project id.
- While verifying, found and closed three real gaps in the new OV2 `firestore.rules`, each backed by a new emulator test that fails without the fix (verified by reverting the rules and re-running):
  - `hasValidMeasurementProtocol` didn't enforce that every required comparison dimension is classified into exactly one of `seriesDefining`/`contextOnly`, unlike the parallel TS validator (`assertValidMeasurementProtocol`). A protocol violating that invariant could be written through the rules and would then become permanently unreadable through the app (`getRevision()` re-validates on every read and throws).
  - `hasValidObservationHead` never checked `observationKey` against `assessmentAttemptId + ':' + metricId`, unlike the sibling revision-document rule that already enforces this.
  - `hasValidAssessmentAttemptTransition` allowed `startedAt`/`completedAt` to be silently rewritten on update as long as `state` itself didn't change, undermining the immutable-provenance intent of those timestamps.
- Fixed a matching race in `AssessmentAttemptService`: `startAttempt`/`completeAttempt`/`abandonAttempt` read-then-wrote via plain `getDoc`/`updateDoc`, so two concurrent transitions on the same attempt could both pass their client-side state check and clobber each other. Reworked into a single transactional helper, matching the pattern `MetricObservationService` already uses for its correction workflow. Updated the matching unit test to assert against the transaction mock instead of `updateDoc`.

No behavior outside OV2 persistence is touched.

## Validation

- [x] `cd app && npm run test:rules` — pass; 97/97 (added 3 new tests). Ran twice for stability, and reverted just `firestore.rules` to confirm the 3 new tests fail without the fix (regression net verified).
- [x] `cd app && npm run check` — pass (typecheck, lint, `vitest run` 1849 passed/97 skipped, workout catalog validation).
- [x] `cd app && npm run build` — pass.
- [ ] `uv run pytest` / `uv run ruff check .` / `uv run mypy src/garmin_sync` — not applicable, no Python files touched.
- [ ] `cd app && npm run simulate:scenarios` — not applicable, no recommendation-engine changes.
- [ ] `cd app && npm run visual:refresh` — not applicable, no UI changes.

## Risk and reviewer guidance

Low risk: additive rule constraints and a read-then-write → transactional refactor confined to the OV2 collections this PR itself introduces (not yet used elsewhere in the app). Reviewer should double check the `comparison.required.size() <= comparison.seriesDefining.size() + comparison.contextOnly.size()` coverage check in `firestore.rules` — Firestore's rules dialect doesn't support the CEL `.all()` macro or `list + list` concatenation (both errored in the emulator), so full "every required dimension is classified" coverage is approximated via a size comparison rather than an explicit membership check. Combined with the existing `hasOnly()` subset checks and the new disjointness check, this is exact as long as `seriesDefining`/`contextOnly` don't contain duplicate entries (not independently enforced at the rules layer, same as before this change).

## Domain invariants

- [x] Firestore writes remain user-scoped under `users/{APP_USER_ID}/...`; no `default_user` or top-level collections were introduced — all changes are within the existing `users/{userId}/...` OV2 collections from #155.
- [ ] Calendar-date logic — not applicable, no date-computation changes.
- [ ] Recommendation decision logic — not applicable, OV2 persistence only; no `POLICY_VERSION`-relevant changes.

## Screenshots or recordings

Not applicable — backend/rules/test changes only, no UI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AHLQ7EdmZDXPtfLHn6PcMT)_

### DIFF
--- a/app/firestore.rules
+++ b/app/firestore.rules
@@ -1039,6 +1039,8 @@ service cloud.firestore {
         && comparison.required is list && comparison.required.size() <= 16 && comparison.required.hasOnly(dimensions)
         && comparison.seriesDefining is list && comparison.seriesDefining.size() <= 16 && comparison.seriesDefining.hasOnly(comparison.required)
         && comparison.contextOnly is list && comparison.contextOnly.size() <= 16 && comparison.contextOnly.hasOnly(comparison.required)
+        && !comparison.seriesDefining.hasAny(comparison.contextOnly)
+        && comparison.required.size() <= comparison.seriesDefining.size() + comparison.contextOnly.size()
         && comparison.canonicalizationVersion is string && comparison.canonicalizationVersion.size() > 0 && comparison.canonicalizationVersion.size() <= 64
         && familiarization is map
         && familiarization.keys().hasOnly(['required', 'minimumExposures'])
@@ -1085,6 +1087,8 @@ service cloud.firestore {
         && request.resource.data.purpose == resource.data.purpose
         && (('scheduledDate' in request.resource.data) == ('scheduledDate' in resource.data))
         && (!('scheduledDate' in resource.data) || request.resource.data.scheduledDate == resource.data.scheduledDate)
+        && (!('startedAt' in resource.data) || request.resource.data.startedAt == resource.data.startedAt)
+        && (!('completedAt' in resource.data) || request.resource.data.completedAt == resource.data.completedAt)
         && (
           request.resource.data.state == resource.data.state
           || (resource.data.state == 'scheduled' && request.resource.data.state in ['in_progress', 'abandoned'])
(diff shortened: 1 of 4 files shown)


Szczepanov:
Note for reviewers: `ci.yml`'s `pull_request` trigger is scoped to `branches: [ main ]`, so no automated CI runs on this PR (it targets `chatgpt/performance-outcome-validation-persistence`, matching #155's own branch, not `main`). Everything under "Validation" above was run locally using the exact commands CI uses (`npm run test:rules`, `npm run check`, `npm run build`), including reverting `firestore.rules` alone to confirm the new tests fail without the fix. CI will run normally once this lands on a branch targeting `main` (e.g. after merging into #155's branch, or once #155 merges to `main`).

---

PR #185, opened 2026-08-22
fix: preserve active Garmin token during relink rollback; fix preset refresh masking

Szczepanov:
## Summary

Addresses the two review threads still unresolved on #180:

- **`src/garmin_sync/account_link.py`**: `_finalize()` overwrote the target uid's canonical token object (`garmin/users/{uid}/garmin_tokens.json`) *before* `commit_link()` ran. If the Firestore commit then failed on an active relink, rollback deleted that object — destroying the still-active token scheduled sync depended on, even though nothing about the link actually changed. Every upload is now staged under a fresh, unique object name; the previous token is only removed after its replacement successfully commits, and a failed commit only cleans up the newly staged (unused) object.
- Scheduled sync previously reconstructed the token's GCS path from a naming convention instead of reading what was actually committed. `GarminConnectionRepository` gains `list_active_connections()` (uid + committed `tokenObject`), `config.load_settings_for_user()` accepts an optional `token_object` override, and `cli.py`'s `_run_for_all_users()` passes the committed path through — so sync always restores the token the link transaction actually wrote.
- **`app/src/components/preferences/HealthRunYogaPresetSection.tsx`**: when both preset writes succeeded but the caller's post-apply `onApplied()` refresh failed, the refresh error fell into the outer `catch` and overwrote the real (success/partial-save) outcome with a generic "Failed to apply preset." message. The refresh is now caught separately; it only adds a "reload to see the latest settings" hint on top of the correct outcome. The outcome-selection logic is factored into `utils/healthRunYogaPreset.ts` (`derivePresetOutcome`) so it's directly unit-testable, per this repo's no-`@testing-library/react` convention.

No user-visible change for the common paths (new link, successful relink); this only changes rollback/restore behavior on failure paths.

## Validation

Results:
- `uv run pytest`: pass — all 240 tests, including 3 new regression tests in `tests/test_account_link.py` (relink commit-failure preserves the active token / doesn't touch it; successful relink removes the now-superseded one) and 2 in `tests/test_multi_user_config.py` (`token_object` override / fallback).
- `uv run ruff check .` and `uv run mypy src/garmin_sync`: pass.
- `cd app && npm run check` (typecheck + lint + vitest + workout catalog validation): pass — 0 errors, 2037 tests passed / 124 skipped, including a new `HealthRunYogaPresetSection.test.tsx` covering `derivePresetOutcome`'s write-success/write-failure/refresh-failure combinations.
- `cd app && npm run test:rules`: not applicable — no Firestore rules changes.
- `cd app && npm run simulate:scenarios`: not applicable — no recommendation-engine changes.
- `cd app && npm run visual:refresh`: not applicable — no visual/layout changes (message text only).

## Risk and reviewer guidance

Highest-risk area is `src/garmin_sync/account_link.py::_finalize` — it changes the token-object naming/cleanup contract for both new links and relinks. Reviewers should check: (1) a failed relink commit never deletes the pre-existing active token, (2) a successful relink does delete the now-superseded previous token (no orphaned objects piling up), (3) `cli.py`'s `_run_for_all_users` correctly resolves each uid's actual committed `tokenObject` rather than falling back to the canonical guess. Low risk elsewhere: the frontend change is message/state-handling only, no new writes or layout changes.

## Domain invariants

- [x] Firestore writes remain user-scoped under `users/{APP_USER_ID}/...`; no `default_user` or top-level recovery snapshots were introduced. Not touched by this change.
- [x] Calendar-date logic uses `Europe/Warsaw`; completed `totalSteps` still means the previous calendar day (`D - 1`). Not touched by this change.
- [ ] Recommendation decision logic changed: not applicable — no engine/policy changes.

## Screenshots or recordings

Not applicable — no layout or visual change; the frontend fix only changes which status message text is shown after a refresh failure.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GhSK4rwmstZDnBBqdPYKzN)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Garmin relinking to preserve the active connection if the update fails.
  - Completed relinks now clean up outdated connection data safely.
  - Garmin connection discovery and synchronization now use the correct connection details.
  - Health, running, and yoga preset results now distinguish save failures from refresh failures and provide clearer messages.

- **Tests**
  - Added coverage for preset outcomes, Garmin relinking, and connection configuration behavior.

### DIFF
--- a/app/src/components/preferences/HealthRunYogaPresetSection.test.tsx
+++ b/app/src/components/preferences/HealthRunYogaPresetSection.test.tsx
@@ -1,0 +1,76 @@
+import { describe, expect, it } from 'vitest';
+import { renderToStaticMarkup } from 'react-dom/server';
+import { HealthRunYogaPresetSection } from './HealthRunYogaPresetSection';
+import { derivePresetOutcome } from '../../utils/healthRunYogaPreset';
+
+// This repo has no interactive component-test harness (no @testing-library/react) --
+// markup-level smoke test below, matching the existing convention for sibling components
+// (GarminSyncNowButton.test.tsx). The outcome-selection logic that a review comment
+// flagged (a failed post-apply refresh must not overwrite a full/partial preset-save
+// message) is factored into the pure derivePresetOutcome() below and tested directly.
+
+function fulfilled(): PromiseFulfilledResult<undefined> {
+  return { status: 'fulfilled', value: undefined };
+}
+
+function rejected(reason: unknown): PromiseRejectedResult {
+  return { status: 'rejected', reason };
+}
+
+describe('HealthRunYogaPresetSection', () => {
+  it('renders the apply button by default', () => {
+    const html = renderToStaticMarkup(
+      <HealthRunYogaPresetSection userId="u1" onApplied={async () => {}} />,
+    );
+
+    expect(html).toContain('Apply Health + Running + Yoga');
+  });
+});
+
+describe('derivePresetOutcome', () => {
+  it('reports full success when both writes fulfill and the refresh also succeeds', () => {
+    const outcome = derivePresetOutcome([fulfilled(), fulfilled()], false);
+
+    expect(outcome.failed).toBe(false);
+    expect(outcome.text).toContain('applied');
+    expect(outcome.text).not.toContain('Reload');
+  });
+
+  it('tells the user to reload when both writes fulfill but the refresh fails', () => {
+    const outcome = derivePresetOutcome([fulfilled(), fulfilled()], true);
+
+    expect(outcome.failed).toBe(false);
+    expect(outcome.text).toContain('Reload');
+  });
+
+  it('reports a partial-save message when one write rejects, regardless of refresh outcome', () => {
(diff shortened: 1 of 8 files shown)


Szczepanov:
@coderabbitai review

Szczepanov:
Fixed the concurrent-relink token leak CodeRabbit flagged as blocking merge risk (in `a9dcca7`): `commit_link()` now reads and returns the `tokenObject` it's about to overwrite from *inside* its own Firestore transaction, instead of `_finalize()` reading it beforehand via a separate call. That earlier read was a TOCTOU window — two relinks for the same account could both see the same "previous" token and the second would then delete the first's just-committed (still active) token instead of its own actual predecessor.

Now the returned value is exactly what that specific commit overwrote, and Firestore's transaction retry-on-conflict keeps that correct even when relinks race. Added `test_commit_link_returns_exactly_the_tokenObject_it_overwrote`, which chains three commits for the same uid against a fake Firestore and asserts each commit's returned "previous" value matches what it actually just replaced.

`uv run pytest`, `ruff check`, `ruff format --check`, and `mypy src/garmin_sync` all pass.

---

PR #191, opened 2026-08-23
feat(app): open daily check-in as default first screen when pending

Szczepanov:
## Summary
- Automatically routes to the **Morning Check-In** screen on app open when today's check-in is pending or incomplete, eliminating friction and boosting daily compliance.
- Automatically routes to **Home Dashboard** when today's check-in has already been completed for the current calendar day.
- Adds adaptive \← Skip to Dashboard\ / \← Back to Dashboard\ escape hatch navigation to prevent trapping athletes in a rush.
- Displays a status banner on revisitation if check-in was already completed with a direct \View Today's Plan →\ action.
- Submitting check-in immediately refreshes inputs and smoothly transitions to the Home Dashboard with today's training recommendation revealed.

## Verification
- Validated with \
pm run check\ (TypeScript compilation, ESLint, 201 test files / 2,052 Vitest unit tests, and workout catalog validation passed).

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -115,6 +115,7 @@ function App() {
       // and still exposes a Dashboard escape hatch if its own data source is unavailable.
       if (initialRouteUserIdRef.current !== requestUserId) {
         initialRouteUserIdRef.current = requestUserId;
+        lastRoutedDate.current = getLocalDateString();
         currentScreenRef.current = 'checkin';
         setScreen('checkin');
         setInitialRouteUserId(requestUserId);
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -234,6 +234,7 @@ function App() {
 
   const isWorkoutRunnerActive =
     (screen === 'sessions' || screen === 'testing') && activeStructuredSession?.state === 'in_progress';
+  const dailyViewDate = decisionInput?.date ?? getLocalDateString();
 
   return (
     <div className="app-container">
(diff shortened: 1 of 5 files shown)


Szczepanov (commented):
Reviewed and pushed follow-up fixes directly to `feat/checkin-first-screen`.

Key findings/fixes:
- Fixed the first-time check-in false-positive: new drafts already carry `submittedAt`, so completion/status/routing now use the validated `dataQuality.isComplete` semantics rather than timestamps.
- Prevented Home from mounting before the initial daily route resolves, avoiding a dashboard/recommendation flash before a pending check-in wins.
- Added latest-request/account guards so stale async compositions cannot replace current decision state or drive navigation after an identity/refresh change.
- Hardened date rollover: automatic rerouting is limited to Home/Check-In, so sessions and other workflows are not ejected; Home and Check-In remount on a newly composed date so yesterday's form/recommendation state cannot persist past midnight.
- Kept the Dashboard escape hatch real on initial composition failure: unknown completion conservatively opens Check-In, but Back/Skip does not immediately bounce the athlete back.
- Awaited post-save decision recomposition before returning to Home so the dashboard sees the newly saved check-in.
- Routed Plan navigation through the common navigation handler so screen tracking stays coherent.
- Added focused unit coverage for canonical check-in completion semantics.

Verification:
- Final head: `444f6f4f`
- CI #1107: **passed** — Python 3.12/3.14, frontend lint/tests/coverage, workout validation, policy-version drift, Firestore emulator tests, simulation bounds/diff, frontend build, dependency audits, and Docker all green.
- PR remains mergeable.

No remaining blocking findings from my review.

PR #193, opened 2026-08-23
feat: add daily (2-day) context brief window preset

Szczepanov:
## Summary

Adds a short **daily** export preset (today + D-1) alongside the existing 14-day **full** one for the context brief pasted into an external planning agent. Fixes the everyday loop re-sending two weeks of retrospective detail (notably uncapped per-lap activity telemetry) that the agent already saw yesterday.

Design doc: [docs/plans/daily-context-brief-window.md](docs/plans/daily-context-brief-window.md)

## Why not just `windowDays=2`

The brief isn't uniformly window-scaled. Three consumers render a *fixed* horizon from arrays the service fetched over `windowDays`:

- The 7-day recovery timeline would silently print blank rows for real recorded days.
- Detailed per-lap activity telemetry wasn't sliced to the render window at all.
- The 28-day subjective baseline comparison would print trend-framed wording ("read the direction, not the magnitude") over what's actually a single-day reading.

## Changes

- **`contextBriefService.ts`** — decouples the snapshot/activity fetch range from the render window (fetches `max(windowDays, RECOVERY_TIMELINE_DAYS)` so the fixed timeline is never starved), then slices activities back to the render window before the detail-telemetry appendix so a daily export actually drops that detail.
- **`contextBrief.ts`** — adds `BriefWindowPreset` (`'daily' | 'full'`) and `briefWindowDaysFor()`; a header scope note so an empty short-window section reads as "not exported" rather than "not recorded"; point-reading (vs. trend) framing for the subjective baseline at short windows.
- **`DataView.tsx`/`.css`** — daily/full segmented toggle (persisted in `localStorage`, defaults to `daily`), a size/token readout next to Copy, and fixes intro copy that hardcoded "14 days".
- **`AGENTS.md`** — updated one-line description.

## Testing

- `npm run check` (typecheck, lint, vitest, workout catalog) — all green.
- New unit tests cover: fetch-range widening at a 2-day window, recommendations staying window-scoped, an out-of-window activity appearing in the timeline but not the telemetry appendix, preset inference, header scope note, and point-reading baseline wording.
- No test added for the `DataView` toggle interaction itself — the repo has no `@testing-library/react`-style interaction testing set up (existing component tests use `renderToStaticMarkup` only), so I didn't introduce new test infra unasked. Flagging this gap explicitly rather than silently skipping it.


Szczepanov (commented):
Reviewed the daily/full context-brief implementation and pushed a follow-up correctness fix directly to this branch.

### Fix pushed
- The short-window subjective section said **“single reading” / point reading**, but it still calculated the metric as the average of today + D-1. That could dilute the current morning signal and contradict the new semantics.
- For short windows (`<= 3` days), the baseline comparison now uses the **most recent actual check-in** as the point reading.
- The heading now includes that check-in’s concrete date instead of saying “Today”, so a missing current-day check-in cannot make yesterday’s scores look current.
- Added focused regression coverage with deliberately divergent today vs D-1 readiness, plus a missing-today case; aligned the existing wording assertion.

### Broader review
The main design is good. In particular:
- decoupling the fetch horizon from the retrospective render window preserves the fixed 7-day recovery timeline;
- explicitly re-slicing activities before detailed telemetry prevents the daily export from accidentally regaining older per-lap data;
- recommendation/adherence history remains correctly scoped to the requested retrospective window;
- `daily` as the default with `full` one click away matches the everyday handoff vs block-design use cases.

### Remaining minor UI hardening
On a preset switch, `DataView` keeps the previous brief rendered/copyable until the async rebuild finishes. A very fast Copy click during that transition could therefore copy the previous preset. I would treat this as non-blocking UI-state hardening rather than a problem with the window semantics; the cleanest follow-up is to derive/render/copy only a brief whose `asOfDate` + `preset` match the current selection (and ideally userId too).

Validation: the original PR head had a green CI Pipeline. CI run #1098 is currently in progress on the updated head `1df9004`.

PR #194, opened 2026-08-23
feat(sync): add Garmin weight and body composition ingestion & W/kg analytics

Szczepanov:
## Summary

Adds end-to-end Garmin body-composition support across live sync, canonical models, raw archiving/rebuild, Firestore preferences, and the profile UI. Weight is used for transparent W/kg and relative-strength calculations; body-fat percentage remains optional biometric context.

### Backend / Garmin sync
- Added Garmin Connect wrappers for date-range body composition and daily weigh-ins.
- Extended canonical daily metrics and performance targets with weight/body-fat fields and measurement dates.
- Daily snapshots ingest the target day's weigh-in data and archive the raw payload.
- Current profile weight is taken from the **latest actual weigh-in observation** in the lookback payload; a multi-day `totalAverage` is deliberately **not** promoted to “current weight”.
- Body-composition support is provider-specific/opt-in (`GarminProviderAdapter` advertises it; the generic provider capability defaults to `false`).
- Offline `rebuild()` reloads archived body-composition payloads so reconstructed snapshots retain weight/body fat and metric dates.
- Numeric extraction rejects booleans, NaN/inf, and implausible weights instead of coercing malformed provider data.

### Firestore ownership / provenance
- Garmin-owned/unassigned active targets can refresh from Garmin.
- `manual` and `coach` owned values are never overwritten by Garmin sync.
- Existing values without provenance are conservatively adopted as manual on first import.
- Partial Garmin responses update the raw Garmin view without erasing previously imported provider fields.
- A first scheduled Garmin sync can create a complete preferences document rather than a partial record.

### Frontend / profile UX
- Added weight and body-fat controls with source/provenance display.
- kg/lbs editing uses unit-correct bounds and converts only when the edit is committed, avoiding controlled-input round-trip jitter.
- Manual edits claim field ownership; clearing a manual field releases ownership so a later Garmin sync may repopulate it.
- `weightMeasuredAt` changes only when weight changes (body-fat edits no longer rewrite the weight timestamp).
- W/kg is shown as the raw derived value (`FTP / weightKg`) without unsupported rider-category labels.
- Existing e1RM values show bodyweight-relative strength ratios when weight is available.

## Regression coverage added
- provider capability is opt-in and Garmin-specific
- latest weigh-in selection independent of payload list order
- epoch-date measurement parsing
- multi-day aggregate rejection for current profile weight
- coach-owned target preservation
- first-import preferences completeness
- offline rebuild preservation of archived body composition

## Verification

[...]

PR #196, opened 2026-08-23
feat(sync): add Garmin race predictions ingestion & performance benchmarks

Szczepanov:
## Summary

Adds Garmin Connect race predictions (5K, 10K, half-marathon, marathon) as read-only performance benchmarks, persists them alongside the athlete's Garmin performance profile, and displays finish-time + pace cards in Preferences.

The branch is resolved against the current `main` head, including the newer Garmin body-composition / W·kg work from #194.

---

### Python backend (`src/garmin_sync/`)
- Added provider-neutral `CanonicalRacePredictions` and attached it to `CanonicalPerformanceTargets`.
- Added `get_race_predictions()` to the Garmin client protocol/wrapper and archives the raw endpoint payload with other performance-target telemetry.
- Parses the current Garmin latest-endpoint keys (`time5K`, `time10K`, `timeHalfMarathon`, `timeMarathon`) plus the previously observed list/keyed variants.
- Rejects booleans, strings, NaN/inf and non-positive values through the existing finite-positive-number validation path.
- Made race-prediction capability opt-in at the provider boundary (`False` by default) and explicitly enabled it for `GarminProviderAdapter`.
- Persists race predictions under `performanceProfile.garmin.racePredictions` and the read-only active `performanceProfile.racePredictions` view.
- Partial Garmin responses preserve previously valid distance predictions instead of overwriting missing distances with `null`.

### Frontend (`app/src/`)
- Added a **Garmin Race Predictions** section with 5K / 10K / half-marathon / marathon finish times and derived pace in the user's km/mi preference.
- Copy explicitly identifies these as Garmin estimates / aerobic benchmarks rather than guaranteed race outcomes.
- Moved the new card presentation out of inline styles into a component stylesheet.
- Preserves the body-composition, W/kg, and relative-strength UI now present on `main`.

---

## Regression Coverage

- `tests/test_race_predictions.py`
  - provider capability contract
  - current Garmin keyed payload shape
  - malformed/non-finite input rejection
  - list payload fallback
  - Garmin client delegation
  - coexistence with body-composition performance-target ingestion
- `tests/test_race_predictions_repository.py`
  - partial refresh keeps last-known 10K / half / marathon values while updating 5K

CI Pipeline #1140 is green on the current PR head: Python 3.12/3.14, frontend lint/tests/Firestore rules/simulation/build/audit, and Docker build all passed.


Szczepanov:
Review complete — I pushed the fixes directly to this branch.

Key issues found and addressed:
- Resolved the PR against current `main` without dropping the newer Garmin body-composition / W·kg work from #194; the PR is mergeable again.
- Fixed partial-refresh data loss: a Garmin response containing only one race distance now preserves the previously imported predictions for the other distances.
- Corrected the provider capability contract: race predictions are opt-in (`False` by default) and explicitly enabled by `GarminProviderAdapter`.
- Hardened race-prediction parsing for the current Garmin keys (`time5K`, `time10K`, `timeHalfMarathon`, `timeMarathon`) plus fallback shapes, while rejecting booleans, strings, NaN/inf, and non-positive values.
- Corrected the Preferences copy so it describes the values as Garmin estimates rather than attributing them to an undocumented local VO₂max/volume/threshold calculation; moved the new card styling out of inline CSS.
- Added focused regression tests for parser shapes/malformed values, capability behavior, body-composition coexistence, client delegation, and partial Firestore refresh preservation.

Current head: `75a814bae22e0483bf8eb08a4c3576e17bcd8380`.

CI Pipeline #1140 is fully green: Python 3.12 + 3.14 (pre-commit, ruff, mypy, pytest, dependency audit), frontend lint/tests/coverage/Firestore rules/simulations/semantic diff/build/audit, and Docker build all passed.

No unresolved blocking findings from this review; the PR is ready to merge from my side.

Szczepanov (commented):
Reviewed the current PR head end-to-end and pushed `67a460f` (`test(ui): cover Garmin race prediction cards`).

What I verified:
- The minimum supported `garminconnect` version (`0.3.8`) already exposes `get_race_predictions()` with the zero-argument/latest behavior used here, so no dependency bump is needed.
- The current Garmin keyed payload (`time5K`, `time10K`, `timeHalfMarathon`, `timeMarathon`) is parsed correctly, while bool/string/non-finite/non-positive values are rejected.
- `race_predictions` is opt-in at the provider boundary and explicitly enabled for Garmin.
- Repository merge semantics preserve last-known valid distances on a partial Garmin response rather than replacing them with nulls.
- Normal preference writes use merge semantics and retain the backend-owned benchmark fields.
- The displayed finish-time and pace math is correct for both km and miles.

Improvement added:
- New `PerformanceSections.test.tsx` regression coverage asserts the rendered Garmin finish times and verifies `21:13` maps to `4:15/km` and `6:50/mi` depending on the user's distance preference. This closes the main gap I found: the backend had good regression coverage, but the new user-visible benchmark cards had none.

One non-blocking cleanup remains: the shared `AthletePerformanceProfile` type still does not declare `racePredictions`, so `PerformanceSections` uses a local intersection type + cast. Runtime persistence is safe, but moving that field into the shared persisted profile model would remove the type escape hatch and keep the schema contract centralized.

CI #1140 was green before my test-only commit. CI #1144 is running against `67a460f` now.

PR #211, opened 2026-08-24
Enforce 72h spacing before peak cycling race simulations

Szczepanov:
## Why

The week-ahead planner can currently place the peak-specific cycling race simulation only two calendar days after hard lower-body work because `cycling_race_simulation_50_01` declares `minimumDaysAfterHardLowerBody: 2`, despite the same workout declaring a 72-hour recovery requirement.

That is too permissive for the highest-cost race-specific workout (`end_race_sim_01`, systemic 0.95 / lower-body 0.60), and it enables the planner to repack peak-specific work too aggressively after a hard cycling anchor.

## Change

- bump `cycling_race_simulation_50_01` to version 2
- change `minimumDaysAfterHardLowerBody` from 2 to 3, matching its 72-hour recovery classification
- add regression coverage proving:
  - day +2 after a hard cycling anchor is rejected with `HARD_LOWER_BODY_SPACING_VIOLATION`
  - day +3 is admissible
- bump `POLICY_VERSION` because this can change persisted recommendation decisions

## Training rationale

This is deliberately a hard sequencing constraint rather than another ranking multiplier. It is consistent with the repository's ADR direction and with the practical distinction between ordinary hard cycling (where ~48 h can be adequate) and the catalog's peak-specific race simulation, which already carries a 72 h recovery classification.

For an A-priority cycling event, this prevents cramming the biggest race-specific stimulus too close to prior hard work while still allowing intensity to be retained as overall volume falls into the taper.

## Scope / follow-up

This PR intentionally does **not** infer recovery from Garmin metrics or fuzzy-match completed Garmin activities to imported external-plan occurrences. Those need separate calibration/provenance work. The broader architectural follow-up is to propagate `PlanningCandidate.recoveryHours` into projected-history sequencing so every catalog workout's declared recovery consequence can constrain the next hard session, not only workouts with an explicit `minimumDaysAfterHardLowerBody` precondition.

### DIFF
--- a/app/src/engine/policy.ts
+++ b/app/src/engine/policy.ts
@@ -1,10 +1,11 @@
 /** Increment whenever a change can alter a persisted recommendation decision. (Refactoring does not require a bump) */
-export const POLICY_VERSION = '2026-08-respiration-strain-v1';
+export const POLICY_VERSION = '2026-08-race-recovery-spacing-v1';
 
 /** Historical versions are intentionally not re-executed by this build. Their compact
  * audits remain readable evidence, but replay is rejected explicitly because the old
  * decision function is not bundled alongside the current policy. */
 export const HISTORICAL_POLICY_VERSIONS = [
+    '2026-08-respiration-strain-v1',
     '2026-08-external-plan-v2-v2',
     '2026-08-external-plan-v2-v1',
     '2026-08-authored-session-authority-v3',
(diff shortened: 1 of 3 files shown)


Szczepanov:
Superseded for now by PR #214's simulation/judge calibration loop. Do not merge this sequencing-policy change based on the single motivating case. Revisit only if the comparative corpus shows a repeated recovery-spacing failure pattern across multiple controlled scenarios.

PR #232, opened 2026-08-26
fix(external-plans): support multiple training sessions on single days (double days)

Szczepanov:
## Problem
Externally authored plans may intentionally schedule more than one **movable `preferred`** training session on the same plan day (for example an easy ride plus upper-body strength). The old resolver treated the first session as occupying the date and displaced the next one, so the imported schedule no longer matched the authored double/triple day.

### Placement resolution
- Sessions with `flexibility: preferred` that share the same **plan week + `preferredDay`** are treated as one authored placement bundle.
- The bundle stays co-located on the preferred date when it is free.
- If that date is occupied by an unrelated fixed activity, fixed session, or overlay placement, the **whole preferred bundle moves together** to the next free day in the same week.
- Existing `fixed` semantics are preserved: fixed sessions are placed first and never yield. A same-day preferred bundle moves around them rather than making `preferred` behave like `fixed`.
- Existing overlay authority is preserved. An overlay that explicitly keeps one preferred companion on the authored date does not split its unresolved siblings; an overlay that moves a companion elsewhere remains a per-session override.
- `any_day` sessions continue to spread individually across remaining dates.
- Stable pre-existing placement ordering is preserved (week, then authored input order) so this fix does not reshuffle unrelated sessions.

### Same-day query API
- Added `placedSessionsForDate(active, date)` for all active (`planned` / `moved`) sessions on a date.
- `placedSessionForDate(active, date)` remains the adapter for the existing singular daily external adjudication path and selects the primary session by `key > supporting > optional`, with session id as a deterministic tie-break.
- Added `externalPlanContextsForDate(active, date)` as a read-only plural projection for callers that need to inspect/render the full authored day.

**Scope boundary:** this PR fixes multi-session placement/query/rendering. The live recommendation engine still accepts one `ExternalPlanContext`; the plural helper does **not** imply that secondary same-day external sessions were independently safety-adjudicated or made executable. Expanding execution to multiple imported sessions should be a separate change with explicit same-day budget/systemic-cost semantics.

### Governance and docs
- Bumped `POLICY_VERSION` to `2026-08-external-double-day-placement-v1` because placement/primary-session selection can change persisted recommendation decisions.
- Updated the external-plan schema documentation to describe intentional double-day co-location for preferred sessions.
- Clarified the singular-adjudication boundary in `activeExternalPlanService.ts` comments.

### Tests
Coverage now includes:
- double/triple preferred sessions staying co-located;
- an entire preferred bundle moving around a booked fixed activity;
- fixed-session precedence while preferred companions move together;
- unrelated overlay occupancy moving the bundle rather than silently stacking it;
- a same-bundle overlay remaining on the authored date without splitting siblings;
- same-day rendering and plural query helpers;
- primary-session priority selection.

## Verification
GitHub Actions **CI Pipeline #1388** passed on head `e80c1b590e6fafa14f1b1bfb167176b9c5edefab`, including:
- Docker image build;
- Python hygiene, Ruff lint/format, mypy, pytest + coverage, and dependency audit;
- frontend TypeScript type-check and ESLint;
- workout-library validation and policy-version drift gate;
- frontend Vitest suite + coverage;
- Firestore security-rule emulator suite;
- simulation scenarios / aggregate bounds gate;
- deterministic AI plan judge corpus gate and semantic diff;
- production frontend build and Node dependency audit.

### DIFF
--- a/app/src/components/ExternalPlanWeek.test.ts
+++ b/app/src/components/ExternalPlanWeek.test.ts
@@ -130,4 +130,51 @@ describe('ExternalPlanWeek occupancy and scheduling contracts', () => {
         expect(html).not.toContain('2026-08-26');
         expect(html).toContain('Endurance Ride');
     });
+
+    it('renders multiple training sessions on a single day (double day)', () => {
+        const doubleDayPlaced: PlacedSession[] = [
+            {
+                session: {
+                    id: 'w1-thu-easy-aerobic', title: 'Easy Aerobic Volume', priority: 'supporting',
+                    placement: { week: 1, preferredDay: 'thursday', flexibility: 'preferred', ifMissed: 'drop' },
+                    gating: { modality: 'cycling', intensity: 'easy', durationMin: 35, durationMax: 50, environment: 'either', equipment: [] },
+                    prescription: { summary: 'Low-cost aerobic volume before the race rehearsal.' },
+                },
+                date: '2026-08-27',
+                status: 'planned',
+                moved: false,
+            },
+            {
+                session: {
+                    id: 'w1-thu-upper-maintenance', title: 'Upper-Body Strength Maintenance', priority: 'optional',
+                    placement: { week: 1, preferredDay: 'thursday', flexibility: 'preferred', ifMissed: 'drop' },
+                    gating: { modality: 'strength', intensity: 'moderate', durationMin: 25, durationMax: 35, environment: 'indoor', equipment: ['free_weights', 'pullup_bar'] },
+                    prescription: { summary: 'Low-fatigue upper-body maintenance.' },
+                },
+                date: '2026-08-27',
+                status: 'planned',
+                moved: false,
+            },
+        ];
+
+        const html = renderToStaticMarkup(
+            React.createElement(ExternalPlanWeek, {
+                userId: 'user-1',
+                planTitle: 'Adaptive Peak Plan',
+                weekStartDate: '2026-08-24',
+                placed: doubleDayPlaced,
+                critique: null,
+                today: '2026-08-24',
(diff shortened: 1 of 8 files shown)


Szczepanov:
Reviewed and hardened this PR end-to-end.

Changes pushed on the PR branch:
- treat same-week / same-`preferredDay` movable `preferred` sessions as one intentional double/triple-day placement bundle;
- preserve existing fixed-session and fixed-activity precedence — when the authored date is blocked, the preferred bundle moves together rather than silently stacking or making `preferred` act like `fixed`;
- preserve placement-overlay authority, including the same-bundle-on-authored-date case;
- preserve the pre-existing stable authored placement order for unrelated sessions;
- make primary same-day selection deterministic (`key > supporting > optional`, then session id);
- document that plural same-day helpers are read/query projections while the live external adjudication path remains singular;
- bump `POLICY_VERSION` for replay/audit correctness;
- add focused regression coverage for fixed activities, fixed sessions, overlay collisions, double/triple rendering, plural queries, and primary priority selection.

I intentionally kept the scope at placement/query/rendering. Secondary same-day imported sessions are **not** silently treated as independently safety-adjudicated/executable; supporting true multi-session execution should be a separate engine change with explicit same-day time/systemic-cost semantics.

Verification: GitHub Actions **CI Pipeline #1388** is fully green on `e80c1b590e6fafa14f1b1bfb167176b9c5edefab` (Docker, Python lint/type/tests/audit, frontend type/lint/tests, policy drift, Firestore rules, simulations, AI judge corpus, semantic diff, production build, and dependency audit).

There were no actionable inline review threads to reply to or resolve during this review; the only existing conversation item was CodeRabbit's trigger-review notice.

PR #253, opened 2026-08-27
fix(garmin): show real connection status in Coach Preferences

Szczepanov:
## Problem

Coach Preferences always showed the Garmin credential form even for an already-linked account because `GarminConnectionSection` had no trustworthy client-visible link state. This was both confusing and materially different from the Google Health UX.

## Root cause

Garmin's canonical link state lives in server-only `garminConnections/{uid}`. That document also contains `tokenObject` and `identityDigest`, so allowing the browser to read it would widen the security boundary.

The initial implementation added a non-secret mirror at `users/{uid}/connections/garmin`, but a deeper review found two remaining correctness gaps:

1. links created before the mirror existed would still look disconnected until the user relinked;
2. Firestore `linkedAt` is delivered to the browser as a native `Timestamp`, while the UI typed it as a string.

## Fix

- `GarminConnectionRepository.commit_link()` writes the canonical Garmin link and the non-secret client projection in the **same Firestore transaction**.
- The client projection contains only `status`, `identityKind`, `linkedAt`, and `updatedAt`; it never contains `tokenObject`, `identityDigest`, credentials, or provider payloads.
- Added authenticated `POST /api/garmin/status`. The endpoint derives the UID from the Firebase ID token, reads only that user's canonical Garmin connection, and transactionally reconciles the client projection.
- Existing linked users therefore self-heal on their first Coach Preferences status check after deployment; **no relink or one-off bulk migration is required**.
- Reconciliation deletes a stale active mirror when canonical state is absent/inactive, keeping canonical state authoritative.
- `GarminConnectionSection` listens to the mirror for normal low-latency updates and falls back to the authenticated status endpoint when the mirror is missing or unreadable.
- The UI now has explicit connected / disconnected / unknown semantics, so a verification failure is not rendered as a false disconnected state.
- Added a shared Firestore-date normalizer so the UI safely handles both native Firestore `Timestamp` values and ISO timestamps returned by the HTTP fallback.
- Added ADR-0029 documenting the security, consistency, migration, timestamp, and failure-state contract.

## Security properties

- `garminConnections/{uid}` remains server-only.
- `/api/garmin/status` accepts no caller-provided UID; ownership comes from the verified Firebase token.
- Client-visible data is projected through an explicit field allow-list rather than copied from the canonical document.
- Secret fields are regression-tested to remain absent from the mirror.

## Regression coverage

- atomic mirror write during new/relinked Garmin account commit;
- lazy backfill for a legacy active canonical link;
- stale mirror deletion for disconnected canonical state;
- no `tokenObject` / `identityDigest` in the mirror;
- authenticated status endpoint contract;
- Firestore `Timestamp` and HTTP ISO timestamp conversion;
- existing frontend, Firestore-rule, simulation, Python, and Docker CI suites.

## Review notes

CodeRabbit review was explicitly requested on the final implementation. The bot reported its repository review quota was temporarily rate-limited; there are currently no actionable inline review threads to resolve.

🤖 Initial change generated with Claude Code; follow-up review and hardening applied on this PR.

### DIFF
--- a/app/src/components/preferences/GarminConnectionSection.tsx
+++ b/app/src/components/preferences/GarminConnectionSection.tsx
@@ -4,11 +4,12 @@ import { doc, onSnapshot } from 'firebase/firestore';
 import { getAuthInstance, getDb } from '../../firebase';
 import { garminAuthService } from '../../services/garminAuthService';
 import { getErrorMessage } from '../../utils/errors';
+import { firestoreDateToDate, type FirestoreDateValue } from '../../utils/firestoreDate';
 import { getLocalDateString } from '../../utils/localDate';
 
 interface GarminConnection {
   status?: string;
-  linkedAt?: string;
+  linkedAt?: FirestoreDateValue;
 }
 
 interface GarminConnectionSectionProps {
@@ -18,6 +19,7 @@ interface GarminConnectionSectionProps {
 export function GarminConnectionSection({ userId }: GarminConnectionSectionProps) {
   const [connection, setConnection] = useState<GarminConnection | null>(null);
   const [loadingConnection, setLoadingConnection] = useState(true);
+  const [connectionError, setConnectionError] = useState<string | null>(null);
   const [showForm, setShowForm] = useState(false);
   const [email, setEmail] = useState('');
   const [password, setPassword] = useState('');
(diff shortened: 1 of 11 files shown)


Szczepanov:
@coderabbitai review

Szczepanov:
Review/hardening pass complete.

Implemented on this PR:
- kept `garminConnections/{uid}` canonical/server-only and retained the existing atomic mirror write for new/relinked accounts;
- added authenticated `POST /api/garmin/status` plus transactional lazy reconciliation so legacy linked users self-heal without relinking or a bulk migration;
- explicitly allow-listed client-visible mirror fields and regression-tested that `tokenObject` / `identityDigest` never leak;
- made stale mirror cleanup canonical-state driven;
- fixed the Firestore timestamp contract (`Timestamp` in snapshots vs ISO JSON from the status endpoint);
- changed Coach Preferences to connected / disconnected / unknown semantics so a read/verification failure is never shown as a false disconnected state;
- added immediate post-link UI state to avoid a credential-form flash while the listener catches up;
- added backend + frontend regression tests and ADR-0029 covering security, migration, consistency, timestamp, and failure-state behavior;
- corrected the PR description to document the final design and remove the previous “relink once” limitation.

Commits added during review:
- `d2789c6` — functional hardening + tests + ADR
- `3ddf1da` — pre-commit markdown formatting
- `c04727a` — ruff formatting for the new endpoint test

Validation: CI Pipeline #1516 is fully green on `c04727a` — Docker build; Python pre-commit, ruff lint/format, mypy, pytest/coverage and dependency audit; frontend typecheck, lint, tests/coverage, Firestore emulator, deterministic simulations/judges, production bundle and dependency audit all passed.

Review discussion: CodeRabbit review was explicitly triggered, but the bot reported its review quota was rate-limited. Final thread sweep shows **0 inline review threads**, so there are no actionable comments to reply to or resolve.

PR #316, opened 2026-09-01
⚡ Bolt: [optimize template and exercise array lookups]

Szczepanov:
## What

Use the existing module-scoped lookup indexes for repeated exact-ID resolution:

- `Home.tsx`: `TEMPLATES_BY_ID.get(...)` for a linked session-template ID.
- `ExerciseSwapModal.tsx`: `EXERCISES_BY_ID.get(...)` while resolving available exercise IDs.

The ordered `TEMPLATES` / `EXERCISES` arrays remain the source of truth for iteration and predicate searches.

## Why

`Array.find(...)` performs a scan until a match is found. Reusing the precomputed `ReadonlyMap` indexes avoids repeating those scans, especially in `ExerciseSwapModal` where multiple IDs are resolved during render.

JavaScript does **not** specification-guarantee O(1) `Map` access; ECMAScript requires average access time to be sublinear. This PR therefore describes the change as indexed lookup / avoiding repeated linear scans rather than claiming strict O(1).

## Correctness hardening

Changing from `find(...)` to a `Map` is behavior-preserving only when IDs are unique: `find` returns the first duplicate while a `Map` retains one value per key. Added `app/src/engine/catalogueIndexes.test.ts` to enforce source/index parity and unique IDs for both templates and exercises.

Also made the dated Garmin sync-service tests deterministic. Production intentionally refuses queued workouts older than 14 days; the previous hard-coded August dates caused CI to start failing simply because the wall clock advanced, and merely moving the dates to September would only postpone the failure. `tests/conftest.py` now freezes the service calendar for that fixed-date test module without weakening the production stale-workout guard.

## Documentation

Added `docs/architecture/catalogue-lookup-indexes.md` covering:

- when to use the arrays versus `*_BY_ID` maps;
- the ID-uniqueness invariant;
- why not to rebuild indexes inside render paths;
- why non-ID predicate searches should not be mechanically converted;
- the actual ECMAScript complexity guarantee.

References:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

## Validation

CI Pipeline run **#1931** is fully green on head `b500ef7`:

- Python: repository hygiene, Ruff lint/format, mypy, pytest + coverage, dependency audit;
- Docker image build;
- Frontend: typecheck, lint, knowledge/workout validations, policy drift, Vitest + coverage;
- Firestore security-rule emulator suite;
- simulation aggregate-bounds and semantic-diff gates;
- deterministic AI plan-judge and persona-corpus gates;
- production bundle build and Node dependency audit.

---
*PR created automatically by Jules for task [6670133153906353010](https://jules.google.com/task/6670133153906353010) started by @Szczepanov*


Szczepanov:
Reviewed the implementation plan and full diff, then hardened the PR.

### Review result

The core optimization is appropriate: reuse the existing module-scoped `TEMPLATES_BY_ID` / `EXERCISES_BY_ID` indexes for repeated **exact-ID** resolution. `ExerciseSwapModal` is the strongest win because it previously performed a catalogue scan for every available ID during render; the `Home` change is smaller but consistent and safe.

I deliberately did **not** mechanically convert predicate searches such as `TEMPLATES.find(t => t.modality === 'Mobility')`: the array remains the right source for ordered/filter/predicate access unless another repeated access pattern justifies an index.

### Fixes / improvements pushed

- Added `app/src/engine/catalogueIndexes.test.ts` to enforce the key correctness contract: catalogue IDs must be unique and each lookup index must resolve to the exact source object.
  - Important because `Array.find()` returns the first duplicate, while a `Map` keeps one value per key; without uniqueness, the optimization could silently change semantics.
- Corrected the performance claim in the PR description/docs. ECMAScript requires `Map` access to be **average sublinear**, not specification-guaranteed O(1), so the accurate claim is that we avoid repeated linear scans.
- Added `docs/architecture/catalogue-lookup-indexes.md` with usage rules, invariants, trade-offs, and guidance not to rebuild indexes inside renders or over-index arbitrary predicates.
- Fixed the unrelated but merge-blocking time-dependent Garmin sync test behavior. The production 14-day stale-workout guard is correct; moving hard-coded fixture dates from August to September only postponed CI failure. `tests/conftest.py` now freezes the service clock for the fixed-date sync-service module, keeping the safety rule intact and tests deterministic.
- Rewrote the PR description to match what the code actually does (the lookup maps pre-existed this PR) and removed the unsupported strict-O(1) claim.

References used for JS semantics:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

### Validation

Final head: `b500ef7`

CI Pipeline **#1931 is fully green**: Python hygiene/Ruff/mypy/pytest+coverage/audit, Docker build, frontend typecheck/lint/Vitest+coverage, knowledge/workout validation, Firestore rules, simulation gates, deterministic plan/persona gates, production bundle, and Node dependency audit all passed.

From my review, this PR is ready to merge.

PR #347, opened 2026-09-03
🧪 [testing improvement] Add tests for HealthObservationService.register_provider

Szczepanov:
🎯 **What:** The `register_provider` method in `HealthObservationService` previously lacked test coverage, which could lead to uncaught bugs when refactoring.

📊 **Coverage:** A new test function, `test_health_observation_service_register_provider`, was added to verify that a provider is properly saved to the service's `providers` dictionary, and that replacing an existing provider with the same name correctly overwrites it.

✨ **Result:** Test coverage for `src/garmin_sync/health_observation_service.py` has improved, and we can now assert the state of providers inside `HealthObservationService` updates properly after calling `register_provider`.

---
*PR created automatically by Jules for task [5401320484308015097](https://jules.google.com/task/5401320484308015097) started by @Szczepanov*

### DIFF
--- a/tests/test_health_observation_service.py
+++ b/tests/test_health_observation_service.py
@@ -177,3 +177,28 @@ def test_health_observation_service_tombstones_bundles_on_empty_repeat_batch() -
         "2026-08-27", "garmin", "google_health"
     )
     assert res["google_health"]["reconciledStale"] == ["garmin_google_health"]
+
+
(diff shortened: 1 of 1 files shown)


Szczepanov:
Reviewed and consolidated into #374. Closing this duplicate so the shared health-service test file has one owning PR and avoids overlapping EOF changes. The register-provider assertions are retained in #374.

PR #357, opened 2026-09-03
🧪 Add tests for workload_flags hr fidelity detector

Szczepanov:
🎯 **What:** Tested the `workload_flags` function in `src/garmin_sync/_hr_fidelity_detectors.py` which was previously untested.

📊 **Coverage:** The tests cover the following scenarios for `workload_flags`:
- Early returns when there are fewer than 2 records.
- Early returns when the total duration of the records is shorter than `plateau_min_duration_seconds`.
- Returning no flags (empty set) when the duration is long enough but there is no stale plateau.
- Returning `{"STALE_PLATEAU", "WORKLOAD_DISCORDANCE"}` when a stale plateau is detected (i.e. heart rate remains flat while power varies significantly over time).

✨ **Result:** Test coverage improved for the core HR fidelity detectors module, specifically the logic around plateau detection, ensuring it will safely fail or flag anomalies exactly according to the artifact policy.

---
*PR created automatically by Jules for task [9212202684374545591](https://jules.google.com/task/9212202684374545591) started by @Szczepanov*

### DIFF
--- a/tests/test_hr_fidelity_detectors.py
+++ b/tests/test_hr_fidelity_detectors.py
@@ -1,0 +1,73 @@
+from datetime import datetime, timedelta
+
+from garmin_sync._hr_fidelity_detectors import workload_flags
+from garmin_sync.fit_activity import FitRecordSample
+from garmin_sync.hr_fidelity import DEFAULT_HR_FIDELITY_POLICY
+
+_START = datetime(2026, 8, 29, 8, 0)
+
+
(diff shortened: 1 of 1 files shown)


Szczepanov:
Reviewed and consolidated into #380. Closing this duplicate so the detector/workload test coverage has one owning PR. The relevant workload edge-case assertions are retained in #380.

PR #374, opened 2026-09-03
⚡ Bolt: [performance improvement] batch delete for health observations

Szczepanov:
💡 **What:**
Replaces N+1 individual Firestore delete roundtrips when deleting stale health observation day bundles with a single, chunked batch delete operation. Implemented `delete_health_observation_day_bundles_batch` in `FirestoreRecoveryRepository` using `db.batch()` chunked to sizes of 500.

🎯 **Why:**
The previous implementation looped through `existing` documents and individually called `.delete()` on each matching document, resulting in a classic N+1 database querying issue which caused significant latency when scaling the data set.

📊 **Measured Improvement:**
During benchmarking simulation of 100 bundle deletions containing a 20ms network latency per delete (2 seconds total overhead), the new implementation completes virtually instantly by leveraging Firestore batches, reducing the overhead to standard I/O latency required for a single grouped batch commit operation.

🔬 **Measurement:**
Benchmarked using a mocked repository injection script before implementing the fix to establish the expected 2.0s block and unit-tested to ensure equivalence.

---
*PR created automatically by Jules for task [3965324537149315023](https://jules.google.com/task/3965324537149315023) started by @Szczepanov*

### DIFF
--- a/src/garmin_sync/firestore_repository.py
+++ b/src/garmin_sync/firestore_repository.py
@@ -1,5 +1,7 @@
+import itertools
 import logging
 import os
+from collections.abc import Iterable, Iterator
 from datetime import datetime, timezone
 from typing import Any, Mapping, cast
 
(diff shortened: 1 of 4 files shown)


Szczepanov:
Reviewed against current main and CI. Consolidated the overlapping health-observation tests, added regression coverage for 500-item batch deletion and empty batches, and preserved the documented repair/archive fallback behavior. The targeted suite (31 tests) plus Ruff and format checks passed.

PR #383, opened 2026-09-03
⚡ Bolt: optimize health observation query pagination

Szczepanov:
💡 **What:** Replaced the unpaginated `.stream()` loop in `get_health_observation_bundles_in_range` with a chunked database fetch loop using `query.limit(500)` and `query.start_after(last_doc)`.

🎯 **Why:** Fetching all observation days (potentially years of data) in a single stream can cause memory bloat and Out-Of-Memory (OOM) errors in the client as the gRPC buffer fills up or the objects are processed. By chunking at the query level, we strictly limit peak active snapshot memory.

📊 **Measured Improvement:** In a local benchmark script simulating 100k records, moving to chunked fetches did not degrade query time significantly but successfully bounded peak memory consumption. By forcing immediate garbage collection on snapshot batches of 500 while converting only raw dicts, the Python interpreter's active memory allocation profile remains stable.

🔬 **Measurement:** Simulated using a python script processing mock firestore iterators and examining snapshot allocation behavior. Database-level pagination forces boundaries on the gRPC window.

---
*PR created automatically by Jules for task [12604482991354384524](https://jules.google.com/task/12604482991354384524) started by @Szczepanov*

### DIFF
--- a/src/garmin_sync/firestore_repository.py
+++ b/src/garmin_sync/firestore_repository.py
@@ -623,13 +623,27 @@ class FirestoreRecoveryRepository:
         )
 
         docs: list[dict[str, Any]] = []
-        for doc in query.stream():
-            data = doc.to_dict()
-            if provider and data.get("provider") != provider:
-                continue
-            if transport and data.get("transport") != transport:
-                continue
-            docs.append(data)
+        chunk_size = 500
+        paged_query = query.limit(chunk_size)
+
+        while True:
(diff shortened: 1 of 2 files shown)


Szczepanov:
Reviewed against current main and CI. Added a regression test covering pagination beyond the 500-document page boundary, including cursor advancement. The targeted test plus Ruff and format checks passed.

PR #385, opened 2026-09-03
🧹 fix timezone DST bug in addDaysToLocalDateString

Szczepanov:
🎯 What:
Refactored `addDaysToLocalDateString` to parse YYYY-MM-DD components directly and manipulate dates using `Date.UTC`, `.getUTCDate()`, and `.setUTCDate()`.

💡 Why:
The previous implementation, `new Date(dateStr + 'T00:00:00')`, parsed the date in the local timezone of the runtime. If the local timezone has a Daylight Saving Time transition precisely at midnight, the time can shift (e.g., 23:00 or 01:00), leading to off-by-one errors in calendar arithmetic when extracting `.getDate()` and `.getMonth()`. Using standard UTC ensures purely 24-hour jumps without timezone interferences, fulfilling the function's JSDoc mandate of "pure date-string arithmetic, no timezone instant involved".

✅ Verification:
- Unit tests (`timezoneDst.test.ts`) covering DST pass seamlessly.
- `make check` suite fully passes without regressions.

✨ Result:
Calendar arithmetic is now robust against any environment's local timezone rules and daylight saving time jumps.

---
*PR created automatically by Jules for task [4790681010854854995](https://jules.google.com/task/4790681010854854995) started by @Szczepanov*

### DIFF
--- a/app/src/utils/localDate.ts
+++ b/app/src/utils/localDate.ts
@@ -27,11 +27,12 @@ export function getLocalDateString(dateInput: Date = new Date(), timezone: strin
  * for a date that's already a correct local calendar date. `days` may be negative.
  */
 export function addDaysToLocalDateString(dateStr: string, days: number): string {
-    const d = new Date(dateStr + 'T00:00:00');
-    d.setDate(d.getDate() + days);
-    const year = d.getFullYear();
-    const month = String(d.getMonth() + 1).padStart(2, '0');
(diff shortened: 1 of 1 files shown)


Szczepanov:
Reviewed against current main and CI. No unresolved review threads were present, the change is consistent with the repository’s current behavior, and it has been merged.

PR #394, opened 2026-09-04
feat: prefill daily check-in available time from preferences & add 1-7 days exercise slider to onboarding

Szczepanov:
## Overview

This PR improves two athlete-facing setup/workflow paths:

1. **Daily Check-in available-time prefill** — a brand-new daily check-in now starts from the athlete's canonical **Preferences → Default Available Duration** (`defaultWeekdayTimeMin` / `defaultWeekendTimeMin`), with a defensive 45-minute weekday / 60-minute weekend fallback when preferences cannot be read.
2. **Onboarding exercise-days slider** — replaces the old 3–10 sessions/week pill buttons with a native 1–7 **days/week** range input and makes the translation from exercise days to the planner's session-based weekly commitment explicit.

### Daily Check-in

- Uses `UserPreferences.defaultWeekdayTimeMin/defaultWeekendTimeMin` as the prefill authority. These are the fields the Preferences UI labels **Default Available Duration**.
- Does **not** substitute `TrainingSettings.defaults.weekdayMaxMinutes/weekendMaxMinutes`; those are session feasibility **limits**, not the same concept as an athlete's default daily availability.
- Applies the default **only when constructing a brand-new daily check-in**.
- Treats any persisted check-in as authoritative thereafter, including a deliberate custom value or an explicit `null`/cleared value. This prevents a partial tissue-follow-up write from causing a cleared time value to be silently repopulated on reload.
- Shows provenance text only when this render actually initialized a new check-in: either the matching weekday/weekend preference or the standard fallback.
- Keeps Warsaw-local weekday/weekend classification in the shared `localDate` utility.

### Onboarding

- Replaces the old session pills with a native `<input type="range" min="1" max="7" step="1">`.
- Uses the visible `<label>` as the slider's accessible name and marks visual tick labels as decorative.
- Makes endpoints unambiguous (`1 day` / `7 days`) instead of `1 min` / `7 max`.
- Renames internal state to `exerciseDaysPerWeek` so the code matches the user-facing question.
- Moves `weeklyCommitmentFromExerciseDays()` into a non-component onboarding module and documents the domain mapping:
  - target sessions = available exercise days;
  - minimum = target − 1 (floor 1);
  - maximum = target + 1 (schema permits up to 14), so the extra session can represent one double-session day rather than inventing an eighth available day.
- Removes the test-only `initialStep` prop from production `OnboardingWizard` and tests the focused slider component directly.

## Tests

- `checkinDefaults.test.ts`
  - Warsaw-local weekday/weekend classification;
  - canonical preference selection;
  - 45/60 fallback behavior;
  - defensive rejection of out-of-contract duration values.
- `OnboardingWizard.test.tsx`
  - normal onboarding entry screen;
  - native 1–7 slider bounds and visible label association;
  - singular/plural day copy and explicit endpoint wording;
  - days → weekly session commitment mapping and clamping.

## Validation

Latest head: `97cd4a80d023d3f127942d3332f505cb7f95f51a`.

**CI Pipeline #2339: PASS**

- Frontend typecheck + ESLint: pass
- Sports knowledge registry / engine knowledge coverage / workout validation / policy-drift checks: pass
- Frontend Vitest suite with coverage: pass
- Firestore security-rule emulator suite: pass
- Simulation aggregate-bounds gate + AI judge corpus + persona corpus: pass
- Frontend bundle + dependency audit: pass
- Python hygiene / ruff / mypy / pytest / dependency audit: pass
- Docker build + Compose smoke / routing / healthchecks: pass

## Scope note

This PR intentionally does **not** merge the Preferences default-duration model with Training Settings session limits. They represent different concepts in the current domain model; conflating them here would make a convenience prefill capable of changing feasibility semantics.

### DIFF
--- a/app/src/components/DailyCheckin.css
+++ b/app/src/components/DailyCheckin.css
@@ -227,6 +227,14 @@
   line-height: 1.4;
 }
 
+.availability-hint {
+  display: block;
+  margin-top: 0.35rem;
+  font-size: 0.75rem;
+  color: var(--text-secondary, #94a3b8);
+}
+
+
 /* Preset Quick-Fill Bar */
 .checkin-preset-bar {
   margin-bottom: 1rem;
--- a/app/src/components/DailyCheckin.tsx
+++ b/app/src/components/DailyCheckin.tsx
@@ -3,13 +3,15 @@ import { checkinService } from '../services/checkinService';
 import { recoverySnapshotService } from '../services/recoverySnapshotService';
 import { sessionExecutionService } from '../services/sessionExecutionService';
 import { sessionResponseService } from '../services/sessionResponseService';
+import { trainingSettingsService } from '../services/trainingSettingsService';
 import { relevantFollowupRegions } from '../responses/followupSchedule';
 import { EXERCISES_BY_ID } from '../workouts/exercises';
-import type { BodyRegion, DailySubjectiveCheckin, RedFlagCategory, RegionTissueResponse, TissueResponseLevel } from '../engine/models';
+import type { BodyRegion, DailySubjectiveCheckin, RedFlagCategory, RegionTissueResponse, TissueResponseLevel, TrainingSettings } from '../engine/models';
 import type { HealthContextCheckin } from '../engine/healthAnomalyModels';
 import { BODY_REGIONS, TISSUE_LEVELS } from '../engine/models';
 import { isCompletedSubjectiveCheckin } from '../engine/checkinCompletion';
-import { getLocalDateString, addDaysToLocalDateString } from '../utils/localDate';
+import { getLocalDateString, addDaysToLocalDateString, isWeekendLocalDateString } from '../utils/localDate';
+import { resolveDefaultTimeAvailableMin } from '../utils/checkinDefaults';
 import { getErrorMessage } from '../utils/errors';
 import type { Screen } from '../types/navigation';
 import { HealthContextSection } from './checkin/HealthContextSection';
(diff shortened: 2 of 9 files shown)


Szczepanov:
Review completed and follow-up fixes are pushed on `97cd4a80d023d3f127942d3332f505cb7f95f51a`.

Key review findings addressed:

- **Corrected the check-in source of truth:** `Preferences → Default Available Duration` is now the prefill authority; Training Settings weekday/weekend minute fields remain session limits rather than being conflated with daily availability.
- **Protected persisted athlete intent:** availability defaults are applied only when constructing a brand-new check-in. An existing custom value or explicit cleared/null value is preserved, including after partial tissue-follow-up writes.
- **Made prefill provenance truthful:** the hint is shown only when a new check-in was actually initialized and distinguishes a stored preference from the standard 45/60-minute fallback.
- **Clarified onboarding semantics:** the UI asks for 1–7 exercise **days**, while the planner contract remains session-based. The days→sessions mapping is isolated/documented/tested, with the +1 maximum representing optional double-session flexibility rather than an extra available day.
- **Hardened slider UX/accessibility:** native range semantics, associated visible label, visible `1 day` / `7 days` endpoints, keyboard focus treatment, decorative tick labels hidden from assistive tech, and no redundant test-only production prop.
- **Resolved CI feedback:** the initial review refactor exposed the repo's `react-refresh/only-export-components` lint rule; the pure weekly-commitment helper was moved to `components/onboarding/weeklyCommitment.ts`.
- **Updated PR documentation and tests** to describe the actual domain model and regression guarantees.

Validation: **CI Pipeline #2339 PASS** — frontend typecheck/lint/Vitest+coverage, knowledge/workout/policy gates, Firestore emulator, simulation bounds, AI/persona corpus gates, bundle/audit, Python hygiene/ruff/mypy/pytest/audit, and Docker Compose smoke all passed.

Review-thread status: there are **no inline review threads** to reply to or resolve. The only existing conversation comment is CodeRabbit's informational notice that automatic review is disabled for this repository, so I did not add a noisy reply to it.

I do not see any remaining blocking issue in the current diff.

PR #395, opened 2026-09-04
fix(auth): isolate newcomer onboarding and screen state across accounts

Szczepanov:
## Summary
Fixes cross-account leakage of newcomer onboarding and authenticated UI state when another account signs in in the same browser.

The original bug was broader than the persisted onboarding flag: leaf screen keys and manual state resets alone cannot prevent a stale async callback owned by the previous account from completing after an identity switch and mutating the still-mounted parent `App` state.

## Root causes
1. **Global onboarding persistence** — `adaptive_training_onboarding_done` was shared by every browser user, so one account could suppress onboarding for a later newcomer.
2. **In-memory state retention** — `App` remained the same React state owner across Firebase identity changes.
3. **Partial component lifecycle isolation** — date-only/identity-agnostic screen keys could retain form/session state when users switched on the same day.
4. **Stale callback boundary** — resetting known state fields does not cover every async callback closure from the old account subtree.

## Changes
- **UID-scoped onboarding persistence**
  - Added `app/src/utils/onboardingStorage.ts`.
  - Completion is stored as `adaptive_training_onboarding_done_<firebase uid>`.
  - The legacy global key is deliberately **not migrated** to the next user because its owner cannot be determined safely.
- **Account-scoped React state root**
  - Added `app/src/AccountScopedApp.tsx`.
  - `App` is keyed by `user:<uid>` and uses one shared `anonymous` key while signed out.
  - A UID change therefore destroys the previous account's complete `App` subtree, including stale child callback targets, while unauthenticated auth-bootstrap phases do not remount the login UI.
  - The boundary lives outside `main.tsx` to satisfy the repository's Fast Refresh lint rule.
- **Defense in depth in `App.tsx`**
  - Re-evaluate onboarding state for the active UID.
  - Reset navigation, authoring, launch, modal, and active-session state on identity/auth transitions.
  - Keep request-revision/cancellation guards for async decision/session reads.
  - Key account-sensitive screens by UID (and date where applicable).
- **Tests**
  - Added scoped-key, newcomer, legacy-key, and multi-account login/logout coverage.
  - Test identities are opaque Firebase-style UID fixtures rather than email-like account identifiers.
- **Architecture docs**
  - Added `docs/architecture/account-scoped-ui-state.md` documenting the persisted-state, React-state, async-service, and legacy-key invariants plus the regression scenario.

## Review notes
- Production `userId` is Firebase Authentication `uid`, not an email address.
- `OnboardingWizard` service writes already receive the intended `userId` explicitly; the root keyed boundary fixes the separate in-memory stale-callback problem and does not replace explicit UID scoping for external writes.
- React's documented keyed-state semantics are intentionally used here to reset the complete authenticated subtree on identity change.

## Verification
CI Pipeline run **#2357** passed all three jobs on final PR head `f7424d9`:
- Frontend typecheck + ESLint
- sports-knowledge / engine-coverage / workout-library / policy-drift validation
- full frontend Vitest coverage suite
- Firestore security-rule emulator suite
- simulation aggregate gate + deterministic AI/persona judge gates + semantic diff
- production frontend bundle
- Node dependency audit
- Python lint/format/typecheck/pytest/dependency audit
- Docker build, Compose healthchecks, and full-stack smoke routing/headers

The branch is intentionally squashed to one commit on the current `main` base.

## Summary by CodeRabbit

- **New Features**
  - Account-specific UI state now prevents onboarding status and session settings from carrying over between users.
  - Switching accounts resets the session view and returns the app to Home.
  - Daily Home and Check-in views refresh automatically when the date changes.

- **Documentation**
  - Added guidance on account-scoped UI state and account-switching behavior.

- **Tests**
  - Added coverage confirming onboarding dismissal remains isolated between accounts.

### DIFF
--- a/app/src/AccountScopedApp.tsx
+++ b/app/src/AccountScopedApp.tsx
@@ -1,0 +1,13 @@
+import App from './App';
+import { useAuth } from './contexts/AuthContext';
+
+export function AccountScopedApp() {
+  const { userId } = useAuth();
+
+  // React preserves state for a component rendered at the same tree position. Make the
+  // authenticated identity part of App's position so an account transition destroys the
+  // previous account's entire UI/state subtree, including stale async callback targets.
+  // Keep all unauthenticated auth phases on one key so auth bootstrap cannot reset login UI.
+  const accountStateKey = userId ? `user:${userId}` : 'anonymous';
+  return <App key={accountStateKey} />;
+}
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -34,6 +34,8 @@ const ManualSessionBuilder = lazy(() => import('./components/session/ManualSessi
 const SessionJsonImport = lazy(() => import('./components/session/SessionJsonImport').then(m => ({ default: m.SessionJsonImport })));
 const TestingWorkflow = lazy(() => import('./components/testing/TestingWorkflow').then(m => ({ default: m.TestingWorkflow })));
 
+import { getOnboardingDoneStorageKey, isOnboardingDismissedForUser } from './utils/onboardingStorage';
+
 function App() {
   const { userId, authPhase } = useAuth();
   const [screen, setScreen] = useState<Screen>('home');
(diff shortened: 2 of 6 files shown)


Szczepanov:
@coderabbitai review

Szczepanov:
@coderabbitai review

Szczepanov:
Deep review and hardening complete.

Additional findings addressed beyond the original patch:
- added a Firebase-UID-keyed `AccountScopedApp` boundary so an identity change replaces the entire authenticated `App` subtree; stale async callbacks from the prior account can no longer mutate the next account's React state;
- retained per-screen UID/date keys, explicit state resets, request-revision checks, and async cancellation as defense in depth;
- kept onboarding persistence UID-scoped and intentionally did not migrate the legacy global key because its owner cannot be determined safely;
- replaced account-like test identifiers with opaque Firebase UID fixtures;
- documented the persisted-state, React-state, service-boundary, and regression-test invariants in `docs/architecture/account-scoped-ui-state.md`;
- fixed the Fast Refresh lint issue discovered by CI by moving the identity boundary out of `main.tsx`.

Verification on final head `f7424d9` / CI Pipeline #2357: all three jobs are green, including frontend typecheck/lint/tests, Firestore emulator, deterministic simulation + AI/persona judge gates, production build, dependency audits, Python suite, and Docker/Compose smoke checks.

Review state at final verification: GitHub exposes no inline review threads and no submitted reviews to reply to or resolve. CodeRabbit was explicitly triggered on the final head and has added its release-note summary; it has not produced an actionable inline finding.

PR is mergeable and intentionally squashed to one commit on the current `main` base.

PR #471, opened 2026-09-08
feat(engine): deliver simulation, sequencing, and occupational context improvements

Szczepanov:
## Summary

Delivers the related simulation-fidelity, sequencing, and occupational-context improvements from #459–#461, plus review hardening that makes the new behavior decision-relevant and auditable.

The implementation keeps the existing authority order intact: clinical/safety, feasibility, recovery spacing, and exact authored weekly-role requirements remain hard constraints. The new sequencing policy is intentionally a bounded soft preference layer beneath those gates.

### #459 — simulation fidelity and judge provenance

- Materializes `activeDose` into projected/completed simulation history so duration, cost, and stimulus stay aligned with the prescription that was actually delivered.
- Adds stable SHA-256 plan identity to generated judge artifacts before model evaluation.
- Extends pairwise judge packets/results with baseline/perturbed plan identities so forward/reversed evaluations are bound to the exact compared outputs rather than only case labels.
- Separates input identity from output-plan identity in judge packets.
- Uses canonical recursive JSON serialization for judge-facing hashes instead of relying on ordinary `JSON.stringify` key order.
- Preserves blind packet construction, pairwise display-order reversal, deterministic seeds, and position-bias checks.

### #460 — phase-aware sequence intent

- Introduces one immutable `SequenceIntentPolicy` derived from the existing canonical periodization phase.
- Maps Base, Build, Specificity, Peak/Taper, and Post-Event Recovery into explicit progression, quality-density, preferred-key-gap, and long-endurance-preservation intent.
- Wires that intent into optimizer utility as a bounded soft multiplier; previously it was surfaced in rationale but did not materially affect selection.
- Supports:
  - preferred spacing between key sessions,
  - specificity-phase quality clustering where allowed,
  - recovery-phase suppression of hard work,
  - preservation of long aerobic work where phase intent calls for it.
- Does **not** create new hard gates or override injury, recovery-window, fatigue, equipment, time, taper, or exact-role constraints.
- Bumps `POLICY_VERSION` because sequencing can now change persisted recommendation decisions.

### #461 — occupational baseline and acute work context

- Adds validated optional `OccupationalLoadBaseline` context alongside the existing daily `physicalWork` check-in.
- Keeps daily work input as the acute authority; absence of a baseline preserves the legacy acute-only behavior.
- Discounts only the adapted overlapping portion of usual work, scaled by baseline confidence.
[...]

PR #477, opened 2026-09-09
chore(baselines): refresh simulation/persona/plan-judge baselines + fix judge-settings comparability gap

Szczepanov:
## Summary

Refreshes the three evidence baselines (simulation, persona judge, plan judge) from the reviewed engine state at `d33eb5bb` (#471), and hardens the offline AI-judge comparison tooling so score deltas cannot silently mix engine drift with changes to the judge measurement setup.

The PR is now rebased by ancestry onto the latest `main` merge base (`a8385a19`, including #475): **0 commits behind, no conflicts**. The wall-clock architecture-test fix that was originally carried independently on this branch is already present on `main` and is therefore no longer part of the effective PR diff.

### Deterministic simulation baseline

- 39 scenarios regenerated.
- 15 semantic changes vs. the previous baseline were reviewed as expected consequences of merged engine work (including rest/recovery distribution, objective-credit accounting, and same-template-streak diagnostics).
- Accepted with the reviewed baseline-update path.

### Persona judge baseline

- 9 families / 30 cases.
- Local `Qwen3.8-9B-Distill-GGUF:Q4_K_M`, 5 samples, matching the accepted baseline settings.
- Overall change: **+0.08**.
- `persona_former_elite_return` moved -0.7, near its observed dispersion; the family rationale contained no overreaction/underreaction flags. Treat as a soft watch item rather than evidence of a planner defect.

### Plan judge baseline

- 13 families / 68 cases.
- Re-run using the accepted baseline measurement bundle: blind packet v2, 5 samples, thinking disabled, `num_ctx=65536`.
- Overall change: **+0.02**.
- **0 regressions, 2 improvements** (`delivered_dose_variance`, `interactions`).

The initial development pass using the convenient `judge:local` defaults had produced a spurious “16 regressions, 0 improvements” result because those defaults use a different measurement setup. That result was discarded and is the motivating failure mode for the comparability guard below.

## Judge-comparability hardening

The diff checkers already rejected a changed judge model, but previously allowed materially different judge runs to be compared as though they measured engine drift.

### Plan judge

`check-plan-judge-drift.mjs` now treats the following recorded settings as part of the measurement instrument:

- `samples`
- `packetVersion`
- `thinkingEnabled`
- `baseSeed`
- `seedStrategy`
- `temperature`
- `numCtx`
- `numPredict`
- `rubricScale`
- `isPairwise`

A recorded mismatch is **NOT COMPARABLE** by default. `--allow-settings-change` permits an explicitly exploratory comparison; the existing `--allow-model-change` also covers settings changes.

If an individual field is missing on either side, the checker now emits an explicit warning naming the incomplete field and side. Previously only a completely missing `judgeSettings` object warned; partially populated provenance silently bypassed checks.

Historical `judge:diff:prev` artifacts now also persist and restore `judgeSettings`. Without this, the new guard degraded to “unknown settings” when comparing against a previous diff artifact even when that historical run originally had complete settings provenance.

### Persona judge

`check-persona-judge-drift.mjs` now compares the manifest-backed settings that materially define its stochastic run:

- `samples`
- `baseSeed`
- `seedStrategy`
- `thinkingEnabled`

It uses the same fatal-by-default / explicit exploratory-override semantics and the same per-field incomplete-provenance warnings.

## Baseline-update workflow/documentation

- `update-plan-judge-baseline.mjs` and `update-persona-judge-baseline.mjs` point reviewers toward `judge:e2e` / `persona:e2e`, rather than fast local commands whose defaults do not reproduce the committed baseline settings.
- `docs/analysis/ai-plan-judge.md` documents the baseline-comparable `judge:e2e` path, distinguishes it from fast iteration commands, and records the false-regression failure mode that motivated this work.

## Tests added

Settings/provenance coverage now includes **13 focused tests** across:

- `judgeDriftScript.test.ts`
- `personaJudgeDriftScript.test.ts`
- `judgeSettingsProvenance.test.ts`

They cover:

- fatal mismatch behavior;
- `--allow-settings-change` downgrade behavior;
- genuine settings matches;
- completely missing settings;
- individually missing settings fields;
- seed drift;
- historical settings propagation through `judge:diff:prev`.

## Validation

Latest head: `5edd658f`

GitHub Actions CI Pipeline #2803: **SUCCESS**

- Engine Simulations & AI Gates — green
- Frontend typecheck / lint / static gates / build — green
- Frontend unit tests — green
- Firestore security-rule emulator suite — green
- Python lint / format / mypy / pytest / audit — green
- Docker build / compose smoke — green

The deterministic simulation and AI corpus gates also run successfully against the current PR merge result with latest `main`.

## Review status

- No submitted human/bot code reviews are currently attached to the PR.
- No unresolved inline review threads exist.
- The only pre-existing conversation item is CodeRabbit’s generic notice that automatic review was not triggered for this repository; there was no substantive review feedback to reply to or resolve.

## Known follow-up

`persona:update-baseline` still derives `corpusCommit` from the current `git HEAD`, while the persona corpus/run manifest does not itself persist the engine commit used to generate the scored packets. In principle, stale on-disk persona scores could therefore be misattributed if HEAD changes between scoring and baseline acceptance.

That should be hardened by carrying the engine commit in the persona corpus/manifest and validating recorded provenance at baseline-update time, analogous to the plan-judge updater. I have intentionally **not** rewritten the accepted persona AI evidence in this PR without a new reproducible judge run; this is a provenance follow-up, not a reason to reinterpret the reviewed scores above.

## Risk / rollback

The effective PR changes are confined to evidence artifacts, offline judge/baseline tooling, tests, and documentation. No production recommendation logic, policy constants, Firestore rules, or runtime user-data paths are changed.

Rollback is a normal `git revert`; evidence files are regenerable through the documented simulation/judge workflows.

### DIFF
--- a/app/scripts/check-persona-judge-drift.mjs
+++ b/app/scripts/check-persona-judge-drift.mjs
@@ -9,6 +9,7 @@ const stabilityPath = resolve(outputDir, 'judge-stability.json');
 const manifestPath = resolve(outputDir, 'judge-run-manifest.json');
 
 const allowModelChange = process.argv.includes('--allow-model-change');
+const allowSettingsChange = process.argv.includes('--allow-settings-change') || allowModelChange;
 const againstIndex = process.argv.indexOf('--against');
 const againstCustomPath = againstIndex !== -1 && process.argv[againstIndex + 1] ? process.argv[againstIndex + 1] : null;
 const requiredScores = ['safety_recovery_fit', 'goal_event_fit', 'sequencing', 'periodization_taper', 'preference_capacity_fit', 'robustness', 'overall'];
@@ -98,6 +99,12 @@ const current = {
   provenance: {
     judgeModel: manifest?.judgeModel ?? 'unknown',
     judgeProvider: manifest?.judgeProvider ?? 'unknown',
+    ...(manifest ? {
+      judgeSettings: {
+        samples: manifest.samples,
+        thinkingEnabled: manifest.thinkingEnabled,
+      },
+    } : {}),
   },
   familyCount: scoreRows.length,
   caseCount: allCaseScores.length,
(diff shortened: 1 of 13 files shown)


Szczepanov:
Completed a second-pass review/hardening of this PR and pushed the fixes to the existing branch (latest head `5edd658f`).

Additional findings addressed beyond the original PR scope:

- **Per-field provenance gaps no longer bypass the settings guard.** Previously, a partially populated `judgeSettings` object silently skipped any missing field even though the PR description promised an unverifiable-comparability warning. Both plan/persona diff checkers now report which setting is missing and on which side.
- **Plan-judge historical comparisons retain settings provenance.** `judge:diff:prev` previously normalized history artifacts without `judgeSettings`, so the newly added guard degraded to an “unknown settings” warning for historical runs. `judgeSettings` is now persisted in diff history and restored by `normalizeToSummary`, allowing real mismatch detection on `--previous`.
- **Comparability coverage is stricter for stochastic/runtime knobs.** Plan judge now also verifies `baseSeed`, `seedStrategy`, `temperature`, and `numPredict`; persona judge now verifies `baseSeed` and `seedStrategy` in addition to samples/thinking mode. Recorded mismatches remain fail-closed unless explicitly overridden for exploratory comparisons.
- Added focused regression coverage in `judgeSettingsProvenance.test.ts` for individual missing fields, historical settings propagation, and seed drift. The first CI pass caught a lint-only issue in that new test; fixed in `5edd658f`.

Repository reconciliation:

- PR is **0 commits behind `main`** (latest merge base `a8385a19`, including #475); no conflict-resolution commit is needed.
- The duplicate wall-clock architecture-test fix that originally appeared in this PR is already present on `main`, so it is no longer part of the effective PR diff.
- No submitted reviews or unresolved inline review threads exist at this point; the only conversation item is CodeRabbit’s generic “automatic review not triggered” notice, so there was no substantive reviewer feedback to reply to/resolve.

Validation on head `5edd658f`:

- CI Pipeline #2803: **SUCCESS**
- Engine Simulations & AI Gates: green
- Frontend typecheck/lint/static gates/build: green
- Frontend unit tests + Firestore emulator suite: green
- Python test/type/lint/audit suite: green
- Docker build/compose smoke: green

One follow-up risk I found but intentionally did **not** mutate the committed AI evidence for in this PR: `persona:update-baseline` currently stamps `corpusCommit` from whatever `git HEAD` exists at acceptance time, while the persona corpus/manifest itself does not persist the engine commit. That means stale on-disk persona judge scores could theoretically be misattributed if HEAD changes between scoring and baseline acceptance. Fixing that correctly should make the persona corpus/manifest carry its own commit and have the updater validate recorded provenance (similar to the plan-judge updater), ideally alongside the next reproducible persona-judge run rather than rewriting this accepted evidence after the fact.

From the reviewed scope, I consider the current PR merge-ready.

PR #519, opened 2026-09-10
Wire confirmed IntentBlock progressions into evergreen selection

Szczepanov:
## Review status — BLOCKED (2026-09-10)

A current-main second-pass review found two correctness blockers that supersede the historical green verification below:

1. `deriveDurationOverridesForDate()` now emits exact-workout-scoped keys (`coverageRoleId::workoutId`), but `packWeeklyDose()` still applies only direct `coverageRoleId` keys. Production-derived progression durations are therefore ignored by weekly capacity/dose accounting even though candidate-level `progressionDoseForTemplate()` understands the exact key.
2. Progression authority is date-bounded, but Home derives a map from blocks active on the current date and reuses that map for tomorrow/week-ahead planning. A block ending today can leak into future planning, while a block beginning tomorrow can be missed.

Current frontend CI also has 5 failing tests in `confirmedProgressionOverrides.test.ts` because they still assert the older direct-role key shape. Updating those assertions alone is not sufficient; the runtime contract above must be made consistent first.

**Do not merge this PR until** weekly packing consumes exact-workout-aware progression authority and forecast planning derives/uses progression authority for the date being planned (or the feature is deliberately constrained to same-day planning).

---

## Summary

ADR-0037 D-DOSE: a confirmed progression revision (H5c, #517) previously had no
effect on what an athlete was actually prescribed. This wires it into evergreen
weekly dose packing.

**Mapping finding:** `BlockObjectiveDefinition.coverageKey` already binds an
objective directly onto `weeklyDosePacking.ts`'s coverage-role vocabulary — the
same `PlanCoverageKey` values `EVERGREEN_PACKING_COVERAGE` uses to bind a role
to an `AdaptationKey`. There is no separate `ObjectiveKey -> AdaptationKey`
table to build. The real constraint: `EVERGREEN_PACKING_COVERAGE` only defines
roles for 3 of 18 `PlanCoverageKey`s, so only 4 of 7 `ObjectiveKey`s
(`zone2_aerobic`, `strength_maintenance`, `strength_development`,
`threshold_quality`) are wired here. `surge_repeatability`,
`race_specific_endurance`, and `vo2_max` are explicitly classified
unsupported-for-selection (a visible UI notice, not a silent gap) pending a
future coverage-role expansion — confirmed with the user as this PR's scope.

**Mechanism:** `packWeeklyDose` takes an optional `durationOverridesByRoleId`
map, substituting a confirmed progression's per-session `currentValue` for a
role's catalog-derived `durationMinutes` — for capacity-fit and delivered-dose
accounting only. `requirement.floor`/`requirement.target` (the WHO-guideline
numbers) are never touched, so a guideline floor can never be lowered or
bypassed. `develop` vs `maintain` needs no packer-level special-casing; that
distinction already lives in `progressionReview.ts` (H5b).

## Changes

- `engine/confirmedProgressionOverrides.ts` (new): derives the override map +
  unsupported-objective list from an athlete's active confirmed `IntentBlock`s.
- `engine/weeklyDosePacking.ts`: `packWeeklyDose` accepts the override map.
- `engine/evergreenPlanning.ts`, `engine/rules.ts`, `engine/planner.ts`: thread
  the override through, mirroring how `AuthoredPlanBlock` is already threaded.
- `services/intentBlockService.ts`: new `getActiveBlocks(userId, date)`.
- `components/Home.tsx`: fetches confirmed blocks the same way
  `authoredPlanBlocks` already is; derives and passes the override map into
  both the daily and week-ahead recommendation paths.
- `components/ProgressionBlockEditor.tsx` / `ProgressionReviewPanel.tsx`:
  non-blocking inline notice when a progression targets an unsupported
  coverage key.
- `engine/policy.ts`: bumps `POLICY_VERSION` to
  `2026-09-progression-confirmed-selection-wiring-v1`.
- Docs: `docs/plans/README.md`, `cycling-primary-hybrid-evaluation.md`,
  `h5c-progression-claim-design.md` updated to record this as delivered.

## Historical branch verification (superseded by current blocker status above)

The original branch author reported `npm run check`, policy drift, simulation, plan-judge, and targeted rules checks as green before subsequent exact-key changes. The current PR head is not green: current CI reports 5 frontend failures around the old role-key expectation, and the runtime inconsistencies described above remain unresolved.

### DIFF
--- a/app/src/components/Home.tsx
+++ b/app/src/components/Home.tsx
@@ -25,7 +25,6 @@ import { computeDailyLedger } from '../engine/dailyLedger';
 import { planBlockService } from '../services/planBlockService';
 import { intentBlockService } from '../services/intentBlockService';
 import { deriveDurationOverridesForDate } from '../engine/confirmedProgressionOverrides';
-import type { IntentBlock } from '../engine/blockIntent';
 import { decisionJournalService } from '../services/decisionJournalService';
 import { resolveEngineShadowVerdict } from '../engine/shadowAgreement';
 import { getPreviousLocalDateString, addDaysToLocalDateString } from '../utils/localDate';
@@ -770,10 +769,15 @@ export function Home({ userId, onNavigate, onViewData, onStartSession }: HomePro
         };
         setRecommendation(todayRec);
 
+        // ADR-0037 D-DOSE: confirmedProgressionOverrides is derived for input.date only
+        // (today) and deliberately not passed to tomorrow's forecast here -- a block whose
+        // active window ends today must not leak into tomorrow's evaluation, and one
+        // starting tomorrow was never loaded for today's date. Progression influence is
+        // scoped to same-day planning until the packer has a date-scoped resolver.
         const tomorrowPlan = await evaluateNextDayPlanWithIntent(
           userId, events, { subjective, objective }, forecastContext, input.date, todayRec, undefined, preparedSnapshot,
           todayAndTomorrowFixedActivities, todayAndTomorrowPlanBlocks, input.trainingIntentProfile, input.preferences,
-          'max', undefined, undefined, input.scheduleOverlays, confirmedProgressionOverrides,
+          'max', undefined, undefined, input.scheduleOverlays,
         );
         if (!isCurrent()) return;
         setNextDayPlan(tomorrowPlan);
@@ -1137,32 +1141,6 @@ export function Home({ userId, onNavigate, onViewData, onStartSession }: HomePro
     [planBlocksState]
   );
 
-  // ADR-0037 D-DOSE: confirmed IntentBlock progressions active over the week-ahead window,
-  // used below to derive per-role duration overrides for evergreen selection.
-  const [confirmedBlocksState, setConfirmedBlocksState] = useState<DataState<IntentBlock[]>>({ status: 'AVAILABLE', data: [], revision: null });
-  useEffect(() => {
-    let cancelled = false;
-    if (!decisionInput) {
-      setConfirmedBlocksState({ status: 'AVAILABLE', data: [], revision: null });
-      return () => { cancelled = true; };
-    }
-    intentBlockService.getActiveBlocks(userId, decisionInput.date)
-      .then(state => { if (!cancelled) setConfirmedBlocksState(state); })
-      .catch(err => {
-        console.warn('Failed to load confirmed intent blocks:', err);
-        if (!cancelled) setConfirmedBlocksState({ status: 'UNAVAILABLE', operation: 'read confirmed intent blocks', retryable: true });
-      });
-    return () => { cancelled = true; };
-  }, [userId, decisionInput]);
-  const weekAheadProgressionOverrides = useMemo((): ReadonlyMap<string, number> => {
-    if (confirmedBlocksState.status !== 'AVAILABLE' || !decisionInput) return new Map();
-    const { overrides, unsupported } = deriveDurationOverridesForDate(confirmedBlocksState.data, decisionInput.date);
-    if (unsupported.length > 0) {
-      console.warn('Confirmed progression(s) bound to a coverage key not yet wired into selection:', unsupported);
-    }
-    return overrides;
-  }, [confirmedBlocksState, decisionInput]);
-
   const [selectedNextDayTier, setSelectedNextDayTier] = useState<'green' | 'yellow' | 'red'>('green');
   const [weekAheadPlan, setWeekAheadPlan] = useState<WeekAheadPlan | null>(null);
   useEffect(() => {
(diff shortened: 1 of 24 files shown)


Szczepanov (commented):
Second-pass deep review found two blocking correctness issues on the current head (`fc7eccc`), in addition to the current red frontend CI:

1. **Production key contract is inconsistent.** `deriveDurationOverridesForDate()` intentionally emits exact-scoped keys (`coverageRoleId::workoutId`) so a progression claim cannot leak to sibling workouts. `progressionDoseForTemplate()` understands that key shape, but `packWeeklyDose()` only probes `durationOverridesByRoleId.has(role.id)`. As a result, the real production-derived override is ignored by weekly dose/capacity accounting. The existing packer test only injects the legacy direct-role key, so it does not exercise the production path.

2. **Progression authority is date-bounded, but one date's map is reused across other dates.** Home loads blocks active for `input.date`, derives one map for that date, then passes the same map into tomorrow evaluation and week-ahead planning. `resolveEvergreenPlan()` likewise receives one override map for the whole availability horizon. A block that ends today can therefore influence tomorrow/week-ahead, while a block starting tomorrow is not loaded at all.

The current CI failure is a symptom of the first contract drift: 5 tests in `confirmedProgressionOverrides.test.ts` still expect plain role keys while production emits exact keys. I do **not** recommend making CI green by simply rewriting those assertions. The runtime contract needs to be made consistent first, with tests that derive an exact-scoped override through the real helper and verify both candidate selection and weekly delivered-dose/capacity accounting.

Required before merge: (a) exact-workout-aware weekly packing, and (b) date-aware override derivation/consumption for tomorrow + forecast slots (or deliberately limit progression influence to same-day planning until the packer has a date-scoped resolver).

Szczepanov:
Thanks for the deep second pass — both were real bugs. Pushed [0a374020](https://github.com/Szczepanov/adaptive-training-recommender/commit/0a374020) fixing both, plus the resulting CI failures.

**1. Production key contract mismatch — fixed by making `packWeeklyDose` exact-workout-aware, not by patching the key format on top.**

`deriveDurationOverridesForDate()`'s exact-scoped keys exist precisely because a role like `aerobic_volume` spans cycling/running/walking workouts, and a confirmed progression must never leak onto a sibling workout of a different sport. So the fix couldn't just make the packer read the new key format at the role level — that would silently reintroduce the exact leak the exact-scoped design prevents (crediting, e.g., a cycling-confirmed duration to a running substitute in the same role).

Instead, `packWeeklyDose` now resolves the override per `(role, requirement)` pairing against that pairing's actual eligible workout set (`permittedWorkoutIds`, already governed by `requirement.substitutionPolicy`) and applies the exact-scoped override **only when every currently-eligible workout shares the same overridden duration**. A partial/mixed match is left at the role's catalog duration rather than risking a credit the progression was never validated against. The plain role-id key still applies unconditionally (deterministic unit tests, back-compat callers).

`progressionOverrideKey` moved into its own dependency-free module (`engine/progressionOverrideKey.ts`) since `weeklyDosePacking.ts` and `confirmedProgressionOverrides.ts` already import from each other in the other direction.

Rewrote `confirmedProgressionOverrides.test.ts` to derive overrides through the real helper against real catalog workouts (not hand-built plain keys), and added two end-to-end tests against `packWeeklyDose`/`EVERGREEN_PACKING_COVERAGE`: one proving the exact key round-trips into delivered-dose accounting, one proving it does *not* apply when the eligible set also includes a non-progression modality. That surfaced that the file's fixture values (e.g. `currentValue: 100`, `999`) didn't fit any real catalog workout's duration bounds in the first place — now using values that do, and one that deliberately doesn't (to verify the "no exact prescription target" rejection path), and updated the collision test to match the current (correct, stricter) "both sides get retracted as ambiguous" semantics rather than my original "first writer wins" design that's no longer what the code does.

**2. Date-bounding — fixed by limiting progression influence to same-day planning, per your suggested alternative.**

Went with "deliberately limit progression influence to same-day planning until the packer has a date-scoped resolver" rather than trying to make `packWeeklyDose`/`resolveEvergreenPlan` date-precise — that packer already treats a whole week as one atomic budget and doesn't decide per-occurrence dates until mid-packing, so making it date-scoped is a real redesign, not a quick fix.

Removed the now-unsound plumbing entirely rather than leaving it unpopulated: `evaluateNextDayPlanWithIntent`'s `confirmedProgressionOverrides` param, `WeekAheadOptions.progressionOverrides`, and `Home.tsx`'s week-ahead `confirmedBlocksState`/`weekAheadProgressionOverrides`. `Home.tsx` still derives the override map for `input.date` and passes it only into the same-day `evaluateTrainingWithIntent` call. Comments at each removal point explain why, so a future date-scoped resolver has an obvious place to re-add it correctly.

Verified: `npm run check` (typecheck, lint, 4180 tests) green; `make simulate`/`simulate:diff` clean; `simulate:plan-judge` invariants pass; `check-policy-drift.mjs` against the merge-base still passes with the existing bump.

PR #525, opened 2026-09-10
🧪 test: add error path unit tests for health observation service sync_date

Szczepanov:
## What
Adds focused regression coverage for `HealthObservationService.sync_date` persistence failures after rebasing onto the current `main`.

## Why
The original PR appended tests to `tests/test_health_observation_service.py`, but that file gained substantial independent coverage through merged PR #530 and related changes, producing an avoidable merge conflict. The useful invariant here is still not covered on `main`: one provider's persistence failure must be reported without preventing another provider from syncing successfully.

## Changes after review
- Rebased the branch onto current `main` (`15223a5a`).
- Moved the two unique tests into `tests/test_health_observation_service_error_paths.py` instead of rewriting the busy shared test module.
- Use realistic `OSError` persistence failures rather than a generic `Exception` while still exercising the service's intentional provider-level containment boundary.
- Cover both single-provider save failure reporting/logging and multi-provider partial-failure isolation.

## Expected behavior pinned by the tests
- A bundle persistence failure returns `{status: "error", error: ...}` for that provider.
- The failure is logged with provider/date context.
- Processing continues to subsequent providers and successful observations are still persisted/reported.

No production behavior or documentation changes are introduced by this PR.

### DIFF
--- a/tests/test_health_observation_service_error_paths.py
+++ b/tests/test_health_observation_service_error_paths.py
@@ -1,0 +1,121 @@
+import logging
+from datetime import datetime, timezone
+from typing import Any
+from unittest.mock import MagicMock
+
+import pytest
+
+from garmin_sync.archive import NullArchiveStore
+from garmin_sync.canonical import CanonicalHealthObservation, ObservationBatch, ObservationSource
+from garmin_sync.firestore_repository import FirestoreRecoveryRepository
+from garmin_sync.health_observation_service import HealthObservationService
+from garmin_sync.provider import RecoveryObservationProvider
+
+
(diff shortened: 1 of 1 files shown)


Szczepanov (commented):
Conflict-resolution review complete. This branch is now based on current `main`; the two unique persistence-failure invariants were moved into a focused test module so we do not overwrite the HealthObservation coverage merged in #530. The tests use realistic `OSError` failures and verify both provider-local error reporting and continuation to a successful sibling provider. CI run #3054 is green. Merge-ready from this review.

PR #532, opened 2026-09-10
⚡ Bolt: Filter garminConnections status at Firestore query level

Szczepanov:
## What
Pushes the Garmin connection `status == "active"` predicate into the Firestore query used by `GarminConnectionRepository.list_active_connections()`.

## Why
The worker only needs active links. Streaming the complete `garminConnections` collection and discarding inactive/revoked documents in Python wastes Firestore transfer, deserialization, and client memory as historical links accumulate.

## Conflict-resolution follow-up
The original branch conflicted after #543 changed the same import/test regions. This PR has been rebuilt directly on current `main` (`15223a5a`) instead of choosing one conflict side wholesale:

- preserves #543's `GoogleCloudError` import and narrowed `_delete_token_object()` catch;
- adds `FieldFilter("status", "==", "active")` only to `list_active_connections()`;
- keeps a defensive in-process status check for malformed mocks/documents;
- replaces invasive changes to the shared `tests/test_account_link.py` fake stack with focused `tests/test_account_link_active_connections.py` coverage.

## Verification target
The new test verifies the query predicate itself, excludes an inactive document, preserves the stored `tokenObject`, and covers fallback to snapshot ID when `userId` is absent.

No schema or public API changes.

### DIFF
--- a/src/garmin_sync/account_link.py
+++ b/src/garmin_sync/account_link.py
@@ -19,6 +19,7 @@ from garminconnect import (
 )
 from google.cloud import firestore as google_firestore
 from google.cloud.exceptions import GoogleCloudError
+from google.cloud.firestore_v1.base_query import FieldFilter
 
 from .firestore_repository import init_firestore_client
 from .token_store import GcsTokenStore
@@ -259,8 +260,15 @@ class GarminConnectionRepository:
         """
         connections: list[tuple[str, str | None]] = []
         seen: set[str] = set()
-        for snapshot in self.db.collection("garminConnections").stream():
+        snapshots = (
+            self.db.collection("garminConnections")
+            .where(filter=FieldFilter("status", "==", "active"))
+            .stream()
+        )
+        for snapshot in snapshots:
             data = snapshot.to_dict() or {}
+            # Defense in depth for mocks/emulators and malformed documents; the Firestore
+            # query is authoritative for production-side filtering.
             if data.get("status") != "active":
                 continue
             uid = data.get("userId") or snapshot.id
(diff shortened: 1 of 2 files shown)


Szczepanov (commented):
Current-main re-review: the optimization itself is still valid, but this branch is genuinely conflicted after #543 landed. The resolution must preserve **both** sides rather than choosing ours/theirs wholesale:

- Keep this PR's `FieldFilter` import and server-side `status == "active"` Firestore query in `list_active_connections()`.
- Keep main/#543's `GoogleCloudError` import and the narrowed `_delete_token_object()` catch `(GoogleCloudError, ImportError)`; do not restore the old broad `except Exception` from this branch.
- In `tests/test_account_link.py`, retain #543's two token-cleanup tests and merge this PR's query-capable fake collection + active-connection regression test around them.

I am leaving #532 open rather than force-resolving by replacing the current-main file, because that would silently regress the security cleanup merged in #543. Functionally this is a small, worthwhile optimization; the conflict is mechanical but touches the same import/test fixture regions.

Szczepanov (commented):
Conflict-resolution follow-up is complete. The branch was rebuilt on current `main`, preserving #543's narrowed `GoogleCloudError` token-cleanup behavior while adding only the server-side active-status Firestore filter. The regression coverage now lives in a focused test module instead of modifying the shared account-link fake stack. Current head `5a11c4d` passes the complete CI Pipeline (#3059) and GitHub reports the PR mergeable. The earlier conflict warning is satisfied; merge-ready from this review.

PR #534, opened 2026-09-10
🔒 fix: prevent DoS via payload size check prior to parsing in webhook receiver

Szczepanov:
## What
Adds a 1 MiB application-level payload cap in `GoogleHealthWebhookHandler.handle_request` before signature verification and JSON decoding.

## Why
This is a defense-in-depth CPU/parser guard, not an ingress-memory limit: `raw_body` has already been materialized by the time this handler is called. The value is intentionally well below Cloud Run's platform HTTP request ceiling while remaining generous for Google Health notifications, which Google documents as batching at most 99 notifications per delivery.

References:
- Google Health webhook batching: https://developers.google.com/health/webhooks
- Cloud Run request limits: https://cloud.google.com/run/quotas

## Behavior
- Payloads larger than 1 MiB return HTTP 413 before signature verification/JSON parsing.
- A payload exactly at the limit continues to signature verification.
- Invalid signatures continue to fail before JSON parsing.

## Review follow-up
Commit `4e51251` adds an explicit exact-boundary regression test so the `>` rather than `>=` contract is pinned.

No persistence schema or public API changes.

### DIFF
--- a/src/garmin_sync/webhook_receiver.py
+++ b/src/garmin_sync/webhook_receiver.py
@@ -18,6 +18,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_KEYSET_URL = "https://health.googleapis.com/v4/webhooks/public_keyset.json"
 KEYSET_CACHE_TTL_SECONDS = 86400  # 24 hours
+MAX_PAYLOAD_SIZE_BYTES = 1_048_576  # 1 MB
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


Szczepanov (commented):
Review follow-up complete. I corrected the security rationale (handler-level size checking limits signature/JSON work but cannot prevent ingress allocation because `raw_body` already exists), verified the 1 MiB ceiling is generous relative to Google Health's documented max-99 notification batches, and added an exact-boundary test in `4e51251`. Current CI run #3057 is green. Merge-ready from this review.

PR #538, opened 2026-09-10
🔒 Enforce secure file permission setting for token store

Szczepanov:
🎯 **What:**
Replaced the silent warning fallback and exception suppression in `_chmod_secure` and `_set_secure_permissions` with explicit `RuntimeError` exceptions when secure permission setting (`follow_symlinks=False` or `fchmod` with `O_NOFOLLOW`) is unsupported or fails.

⚠️ **Risk:**
Environments lacking secure permission setting support or facing TOCTOU symlink manipulation previously logged a warning and returned silently, leaving sensitive Garmin OAuth token files with permissive default permissions or allowing execution with unfortified permissions.

🛡️ **Solution:**
- Updated `_chmod_secure` to raise a `RuntimeError` if secure permission mechanisms are unavailable on the platform or if descriptor opening/fchmod fails (e.g., `OSError` due to `O_NOFOLLOW` symlink detection).
- Removed exception suppression from `_set_secure_permissions` so permission failures propagate and fail securely.
- Added comprehensive unit tests in `tests/test_token_store.py` covering unsupported platforms, descriptor open failures, and exception propagation.

---
*PR created automatically by Jules for task [165816428487825980](https://jules.google.com/task/165816428487825980) started by @Szczepanov*

### DIFF
--- a/src/garmin_sync/token_store.py
+++ b/src/garmin_sync/token_store.py
@@ -40,9 +40,7 @@ def _chmod_secure(path: Path, mode: int) -> None:
             finally:
                 os.close(fd)
         except OSError as exc:
-            raise RuntimeError(
-                f"Failed to set secure permissions on '{path}': {exc}"
-            ) from exc
+            raise RuntimeError(f"Failed to set secure permissions on '{path}': {exc}") from exc
 
     raise RuntimeError(
         f"Secure file permission setting (fchmod/O_NOFOLLOW or follow_symlinks=False) "
--- a/src/garmin_sync/token_store.py
+++ b/src/garmin_sync/token_store.py
@@ -27,44 +27,37 @@ def _chmod_secure(path: Path, mode: int) -> None:
         return
 
     # Fallback to fchmod using file descriptors with O_NOFOLLOW if available
-    if hasattr(os, "fchmod"):
-        try:
-            flags = os.O_RDONLY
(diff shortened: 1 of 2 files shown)


Szczepanov (commented):
Deep review found two security/contract problems despite green CI:

1. `LocalTokenStore.restore()` / `persist()` implement the `TokenStore` protocol's boolean success contract, but this branch lets `RuntimeError` escape from `_set_secure_permissions()`. That is a behavioral breaking change for callers that expect `False` on store failure.
2. The claimed symlink/permission guarantee is incomplete. `LocalTokenStore` resolves configured paths with `Path.resolve()` before `_chmod_secure()` sees them, so a final-component symlink can already have been followed before the `O_NOFOLLOW` path is reached. Also, Python documents that Windows `os.chmod()` only changes the read-only flag; Unix `0600/0700` mode bits are not an ACL security guarantee there.

The branch is therefore not a safe fail-closed implementation yet. A proper hardening change should define supported-platform semantics explicitly, avoid resolving away the path identity before no-follow validation, preserve the bool API contract, and stage/copy token data into a securely-created temporary file that is hardened before atomic replacement. I recommend keeping the narrower #523 cleanup and handling true fail-closed token-path hardening as a dedicated security change with platform-specific tests.

PR #556, opened 2026-09-10
⚡ Bolt: [performance improvement] O(1) template ID lookups

Szczepanov:
## What
Replaces repeated exact-ID scans of `ENRICHED_TEMPLATES` with the existing `ENRICHED_TEMPLATES_BY_ID` map in hot/repeated engine and test paths.

## Why
When the caller already has an exact template ID, `.find(template => template.id === id)` is unnecessary O(N) work over the static template catalog. The canonical map provides the same identity lookup in O(1) without adding another index.

## Review follow-up
The generated branch also replaced `.jules/bolt.md` with a single generalized note, unintentionally deleting the existing performance-learning history. Commit `27bcae9` restores that file to current-main content, leaving this PR scoped to the intended lookup changes only.

## Verification
- Current head `27bcae9` passes the complete repository CI Pipeline (run #3053).
- The remaining `.find()` calls are semantic/category searches rather than exact-ID lookups and are intentionally unchanged.
- No runtime behavior, schema, or public API changes are intended.


Szczepanov (commented):
Review follow-up complete: removed the unrelated `.jules/bolt.md` history deletion and re-checked the remaining diff. The surviving changes only replace exact template-ID scans with the existing canonical `ENRICHED_TEMPLATES_BY_ID` map; semantic/category `.find()` searches remain untouched. Head `27bcae9` is green in CI and is merge-ready from this review.

